PR #29, opened 2019-12-09
Include tests in sdists

toddrme2178:
This makes it easier for downstream packagers to make sure they have the package set up correctly

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,9 @@
-include README.rst
-include examples/*.py
-include acoular/xml/*.*
 include LICENSE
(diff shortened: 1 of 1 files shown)


esarradj:
The Windows builds are failing on travis with this. Not sure why this is the case ...

toddrme2178:
It looks like there is a problem with the traits install during the build.  I am not sure it has anything to do with this commit.

esarradj:
Without this commit travis on Windows builds just fine.
see https://travis-ci.org/acoular/acoular and https://travis-ci.org/acoular/acoular/builds/622744559

PR #42, opened 2020-12-03
Switching from arange to linspace in SineGenerator

gomeslucasm:
I identified one error in creating the time vector in the SineGenerator method. To create the time vector is used the arange method instead linspace, and because of that, the generated signal will be not a clean sine (or pure tone).

The comparisons using linspace and arange can be seen in this notebook from google colab: https://colab.research.google.com/drive/1fflSEMuR46BnDc1iX0hZZ0OqknWK6uVZ?usp=sharing

I plotted the graphics in the frequency domain and in them you can see that the use of the arange generates several unwanted artifacts.

### DIFF
--- a/acoular/signals.py
+++ b/acoular/signals.py
@@ -18,7 +18,7 @@
 
 # imports from other packages
 from __future__ import print_function, division
-from numpy import pi, arange, sin, sqrt, repeat, tile, log, zeros
+from numpy import pi, sin, sqrt, repeat, tile, log, zeros, linspace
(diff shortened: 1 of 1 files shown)


esarradj:
I've checked the issue and found no problem. The differences you found are just due to the fact that arange does not include the max value and linspace does. If you use 

t = np.linspace(0,tmax-tmax/fs,int(tmax*fs))

in the colab notebook instead, both results are the same, as expected.

PR #62, opened 2022-09-28
Remove deprecated function has_key

boneyag:
`has_key` is deprecated since Python 3 in favour of `in` operator.


esarradj:
Thank you very much for pointing this out. However, fileimport.py is depreciated and no longer maintained or tested, so it shouldn't be changed.

PR #72, opened 2023-04-03
Remove import of np.int

bnavigator:
`np.int()` has been removed in NumPy 1.24 and was just an alias to builtin `int()`


esarradj:
Thanks for pointing this out. It is meanwhile resolved in 13f78f5aeb2fbff6904ae9ad663927e140f1ae89.

Issue #209, opened 2024-04-09
Github CI test pipeline sporadically fails for macos


adku1173:
Error:
`run_tests.sh: line 8:  1871 Illegal instruction: 4  python -m unittest discover -v -p "test_*.py"
Error: Process completed with exit code 132.`

However, rerunning the pipeline resolves the problem.

adku1173:
If the workflow is adjusted to use `run_tests_osx.sh`, the same problem occurs with python 3.11 
`run_tests_osx.sh: line 8:  1959 Illegal instruction: 4  python3 -m unittest discover -v -p "test_*.py"`

adku1173:
After some logging with the existing test pipeline and different Mac OS versions, the problem seems to be the CPU's FMA feature's availability. 

If we compare the two workflow runs for the same macOS image and Python version (3.10):

ProductVersion:	12.7.4
BuildVersion:	21H1123
i386

* [Run1 (failed)](https://github.com/adku1173/acoular/actions/runs/8735322584/job/23967868387)

  -  FMA not available (CPU: Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz)

* [Run 2 (succeeded)](https://github.com/adku1173/acoular/actions/runs/8735500283/job/23968382509)

  - FMA available (CPU: Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz)

As we use SIMD instructions via Numba, this presumably causes the problem.

esarradj:
Numba does autodetect (?) the CPU features, see [here](https://numba.pydata.org/numba-doc/dev/reference/envvars.html), so I have some doubts that it is to blame

esarradj:
The choice of hardware for macos-latest is nondeterministic (see [here](https://github.com/actions/runner-images/issues/8642#issuecomment-1824251689) ) and the problem may go away eventually, when macos-latetest refers not to macos-12, but macos-14 .  In the meantime, it says [here](https://github.com/bancaditalia/black-it/pull/77/commits/5c22644a50184c24b690fc42fe5a7a172163ddd2) that pytables may the problem. However, the workaround seems to be a bit too expensive for us.

How about explicitly requesting macos-13 or macos-14 instead of macos -latest which gives us 12.7 ?

adku1173:
[...]

PR #217, opened 2024-04-17
update zenodo.json

adku1173:
* Updates the current zenodo.json data
* validated build can be found here: https://sandbox.zenodo.org/records/46339

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -4,7 +4,7 @@ name: Acoular install and test
 
 on:
   push:
-    branches: [ "*" ]
+    branches: [ master ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -37,7 +37,7 @@ jobs:
         cd ../acoular/acoular/tests/
         bash run_tests.sh
       if: ${{ matrix.os == 'ubuntu-latest'}} 
-    - name: Test with unittest *nix
+    - name: Test macos with unittest *nix
       run: |
         sysctl -a machdep.cpu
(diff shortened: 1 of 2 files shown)


esarradj (commented):
Test for MacOS Python 3.11 fails.

adku1173:
As already mentioned in #209, the test pipeline fails sometimes. In this case, rerunning does not solve the problem. I will look into the issue soon.

.github/workflows/python-package.yml
```diff
@@ -30,14 +30,23 @@ jobs:
...
         pip install .
         python -c 'import acoular'
+    - name: Test ubuntu with unittest *nix
+      run: |
+        cat /proc/cpuinfo
+        chmod +x ../acoular/acoular/tests/run_tests.sh
+        cd ../acoular/acoular/tests/
+        bash run_tests.sh
+      if: ${{ matrix.os == 'ubuntu-latest'}} 
     - name: Test with unittest *nix
```
esarradj, line 40: Isn't this macos ?

adku1173: That's obviously ubuntu.  Why should it be macos? See the latest test action: https://github.com/acoular/acoular/actions/runs/8736226608

esarradj: because then we have already Test ubuntu with unittest *nix in l 33
and now it is Test with unittest *nix in l 40 when actually macos-latest is requested

I think the '*nix' was a label for both ubuntu and macos and now we have ubuntu and macos as different actions. So they should have correct labels

adku1173: What is this '*nix' label for?

esarradj: [Li,U]nix

esarradj (approved):

PR #222, opened 2024-04-21
Add code of conduct

artpelling:
We need to add a contact email for reports. Do we have one already?

In order to encourage a report, it could be good to nominate two designated people (male and female) to report to and supply their personal email addresses. A general address like `info@acoular.org` or similar that gets forwarded to all main developers might be discouraging if the report concerns one of the devs.


adku1173:
The email address is indeed `info@acoular.org` ;)

adku1173 (approved):
:+1:

PR #223, opened 2024-04-21
Add CITATION.cff

artpelling:
This adds a citation cff file that should display nicely in the sidebar. I think that this can also make `zenodo.json` obsolete. Maybe we can test this also with the sandbox. 

Do we have a release checklist? The release date, dois and version need to be updated here when a new release is made.


adku1173:
There is a release checklist on our internal wiki at TU Berlin. Maybe we should move this information to the GitHub Wiki? 

I am confused, because the checkbox on the GitHub page does not show the expected APA and Bibtex tabs. Only the .cff file is linked. 

image

adku1173:
I already tested it with the Zenodo sandbox on my private fork. However, one would need to remove the .zenodo.json file, as described in: https://help.zenodo.org/faq/ , which I don't think is necessary.

esarradj:
I'd say we should either have the authors listed in zenodo.json OR in citation.cff to avoid conflicting metadata. With citation.cff however we cannot guess the DOI of an upcoming release, can we?

Would it be an option to put just the zenodo citation into citation.cff and control everything using zenodo.json?

Another option could be to trust zenodo handling the citation.cff correctly and have all the data except the zenodo DOI there. 

Btw. I found that the APA / BibTeX citations offered will show the preferred-citation data within citation.cff, in our case the Applied Acoustics article from 2017. Do we want to have it this way?

adku1173:
According to the zenodo FAQ, it is possible to know the DOI beforehand :

_On the upload page under Basic Information and Digital Object Identifier click the Reserve DOI button. The text field above will display the DOI that your record will have once it is published. This will not register the DOI yet, nor will it publish your record (so you can still update the files). This DOI can be safely used in the record's own content as well as any other separate datasets or papers you might be planning to publish._

artpelling:
> I'd say we should either have the authors listed in zenodo.json OR in citation.cff to avoid conflicting metadata.

I agree. There should be only one location where this information is supplied to keep things maintainable. Seeing that cff is a more general format, I would prefer it over something platform-specific.

>  With citation.cff however we cannot guess the DOI of an upcoming release, can we?

I think this is possible. As @adku1173 pointed out, https://help.zenodo.org/faq/ suggests that you can use Zenodo's reserve DOI feature for this purpose.

> Would it be an option to put just the zenodo citation into citation.cff and control everything using zenodo.json?

I think that would be very incovenient.

> Another option could be to trust zenodo handling the citation.cff correctly and have all the data except the zenodo DOI there.

IMO this is the way to go. If the zenodo reserve DOI feature works as expected, there should be no drawback.

> Btw. I found that the APA / BibTeX citations offered will show the preferred-citation data within citation.cff, in our case the Applied Acoustics article from 2017. Do we want to have it this way?i

Being cited by a journal article feels like the "proper" way to do it. I suggest to add a message field to `CITATION.cff` that  and also add a small comment in the `README.md` that advises to cite both the software and the journal article.

artpelling:
> There is a release checklist on our internal wiki at TU Berlin. Maybe we should move this information to the GitHub Wiki?
> 
I think its a good idea to move it to gh. However, the release checklist might contain sensitive information. We should check for this. Maybe we can find a place on gh that restricts access only for main devs.

> I am confused, because the checkbox on the GitHub page does not show the expected APA and Bibtex tabs. Only the .cff file is linked.

I had a typo in the file that I fixed. Now it shows up correctly. It does so without `preferred-citation`, only `reference` is enough which I like.

adku1173:
I think removing .zenodo.json is not an option. I tested both variants with the zenodo sandbox:

* no zenodo.json: https://sandbox.zenodo.org/records/47715
* with zenodo.json: https://sandbox.zenodo.org/records/47715

So far, using only CITATION.cff will, of course, not lead to a sufficient description and title.
Let's see if there is a way to achieve the same results with the .cff file. The title can be changed to: `Acoular – Acoustic testing and source mapping software`.

artpelling:
> I tested both variants with the zenodo sandbox:
> 
>     * no zenodo.json: https://sandbox.zenodo.org/records/47715
> 
>     * with zenodo.json: https://sandbox.zenodo.org/records/47718
> 
> 
> So far, using only CITATION.cff will, of course, not lead to a sufficient description and title. Let's see if there is a way to achieve the same results with the .cff file. The title can be changed to: `Acoular – Acoustic testing and source mapping software`.

I've changed the title and added an abstract. Can you check how it is rendered now?
By default, with a cff file, zenodo takes the repo description as description. This is rather short, so a longer version should probably be in the form of an abstract. I've notced that the github description "Library for acoustic beamforming" does not match the subtitle of "Acoustic testing and source mapping software". Should we adjust this? 
Also the github homepage link uses `http://www.acoular.org`, maybe `https://acoular.org` is nicer...

artpelling:
If the HTML in the `abstract` field is not rendered correctly by zenodo, we could maybe omit the highlights list as a compromise.

adku1173:
> If the HTML in the `abstract` field is not rendered correctly by zenodo, we could maybe omit the highlights list as a compromise.

It is indeed rendered correctly: https://sandbox.zenodo.org/records/47723

Thus, we can remove the zenodo.json metadata file :)

artpelling:
OK I think this is ready now! @adku1173 @esarradj

adku1173 (approved):

Issue #215, opened 2024-04-09
Implement linting


adku1173:
- [ ] decide on appropriate linting settings (and linter)
- [ ] add github workflow that adds linting as an additional test

artpelling:
Closed by #232

PR #229, opened 2024-04-22
Basic Linting

### LINKED ISSUE #213: Remove pylint codes from file headers
disable certain codes will be handled in the pyproject.toml file

```
#pylint: disable-msg=E0611, E1101, C0103, R0901, R0902, R0903, R0904, W0232
```

adku1173:
This PR introduces Ruff as the linting tool and a corresponding GitHub workflow. Due to the linting milestone and its associated issues, many of the rules are currently ignored. This PR already fixed some rules (automatically), such as white spaces, formatting, etc. 

fixes #213, #215


pyproject.toml
```diff
@@ -56,6 +56,7 @@ dev = [
     "sounddevice", # docs 
     "graphviz", #docs
     "pickleshare", #docs
+    "ruff>=0.3.5",
```
artpelling, line 59: I think it would be good to pin the version in order to ensure that CI is matching local ruff perfectly.
I would also suggest to use the latest version `0.4.1` which implements LSP.

.ruff.toml
```diff
@@ -0,0 +1,130 @@
...
+[lint.per-file-ignores]
+"acoular/__init__.py" = ["F401"] # unused import -> TBD
+"acoular/demo/__init__.py" = ["F401"] # unused import -> TBD
+"acoular/configuration.py" = ["F401"] # unused import (will be solved by issue #212)
+"acoular/nidaqimport.py" = ["F841"] # unused variable (really unused?) -> TBD
+"acoular/tbeamform.py" = ["F841"] # unused variable (really unused?) -> TBD
+"acoular/fbeamform.py" = ["W605"] # invalid escape sequence -> TBD
+"acoular/tests/test_tprocess.py" = ["F403"] # star import 
+"acoular/tests/*" = ["D"] # exclude docstring linting
+"acoular/version.py" = ["D"] # exclude docstring linting
```
artpelling, line 130: Do all of these concern multiple errors or only some instances in these files? If the latter is the case, it would be probably better to use `# noqa: ["X000"]` directives.

.ruff.toml
```diff
@@ -0,0 +1,130 @@
+# Exclude a variety of commonly ignored directories.
```
artpelling, line 1: This whole config belongs in `pyproject.toml` under `[tool.ruff]` IMO.

.ruff.toml
```diff
@@ -0,0 +1,130 @@
...
+    ".vscode",
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "site-packages",
+    "venv",
+]
```
artpelling, line 30:
```suggestion
src = ["acoular"]
exclude = ["acoular/tests/unsupported"]
```

artpelling (changes requested):

.ruff.toml
adku1173: I don't want to clutter the pyproject.toml file. I would rather stick to a separate ruff.toml, if this is not a problem for someone

pyproject.toml
adku1173: Indeed.

.ruff.toml
artpelling: For me, the appeal of `pyproject.toml` is that it eliminates most other config files and has everything in a central location. I don't think that it will become cluttered. We could just add it to the end of the file. As hopefully more and more linting excludes are removed it should also shrink in the long run.

artpelling: OK I've changed my mind. It might actually be better this way.

artpelling (approved):
@adku1173 Could you link/resolve any related issues that are affected/(partially) solved by this PR?

adku1173:
> @adku1173 Could you link/resolve any related issues that are affected/(partially) solved by this PR?

Done! It's #215 and #213  (see PR header)

artpelling:
@gherold Ready from my side :green_circle:

PR #232, opened 2024-04-22
Hatch based tests

artpelling:
This PR will use `hatch` for CI. This should simplify the tests substantially and centralize the CI config in `pyproject.toml`


artpelling:
@esarradj @adku1173 Aside from the broken links in the Markdown, this should be ready from my side.

esarradj:
The link problem is about some deep links to the acoular.org site itself using http instead of https. This stopped working when we actually moved the domain to GitHub serving the docs.

esarradj (approved):

artpelling:
I've now also added a coverage report to the CI. We are at 56% currently.

We could think about whether we want to use something like codecov to analyze and report the coverage statistics and add a badge. But I don't see the immediate need and won't do that in this PR. 

I've also updated the CI badge in the readme. Now that all tests pass, I think we are ready to merge. I won't add anything anymore in this PR.

acoular/demo/__init__.py
```diff
@@ -1,7 +1,7 @@
 # coding=UTF-8
```
adku1173, line 1:
```suggestion
```

By default, Python source files are treated as encoded in UTF-8. (https://docs.python.org/3/tutorial/interpreter.html#source-code-encoding)

acoular/internal.py
```diff
@@ -1,27 +1,26 @@
 # coding=UTF-8
```
adku1173, line 1:
```suggestion
```

acoular/version.py
```diff
@@ -1,9 +1,9 @@
 # coding=UTF-8
```
adku1173, line 1:
```suggestion
```

acoular/__init__.py
```diff
@@ -1,7 +1,7 @@
 # coding=UTF-8
```
adku1173, line 1:
```suggestion
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,73 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: build-docs
+      - name: Upload docs to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v4
+        if: ${{ github.event_name =='workflow_dispatch' && github.ref == 'refs/heads/master' }}
+        with:
+          publish_branch: gh-pages
```
adku1173, line 46: I think it would be better to move the `if: ${{ github.event_name =='workflow_dispatch' && github.ref == 'refs/heads/master' }}` statement directly under the `deploy` job statement so that the `deploy` job is skipped (in case it was not triggered manually). Currently, the deploy job is executed, although deployment will never be done since it needs to be triggered manually via the `workflow_dispatch` action.  

If we let it as it is, the job output looks like the docs have been deployed:
image

But in fact, the deployment (upload to GitHub pages) was skipped: 
image

adku1173 (changes requested):

acoular/demo/__init__.py
artpelling: I didn't add this. I suggest to "unignore" `UP012` and let ruff handle this.

.github/workflows/docs.yml
artpelling: This was intentional to check if the artifacts work. But I see how this could be misleading. It is probably better to skip the enitre job.

acoular/demo/__init__.py
artpelling: I just confused `UP009` with `UP012`. I will remove the encoding statements but will also unignore `UP012` while at it.

adku1173 (approved):

PR #236, opened 2024-04-25
Draft: Use Context Manager

adku1173:
I would like to draft some changes regarding the handling of files. Currently, the `TimeSamples` class (and other classes) holds an open file via the `h5f` file trait. This becomes a problem when the same file should be overwritten or changed by some other class (e.g. the `WriteH5` class) but the `TimeSamples` instance still exists. When running the acoular_demo.py twice in an IDE would result in the error:  

```
HDF5ExtError: HDF5 error back trace

  File "H5F.c", line 660, in H5Fcreate
    unable to synchronously create file
  File "H5F.c", line 614, in H5F__create_api_common
    unable to create file
  File "H5VLcallback.c", line 3605, in H5VL_file_create
    file create failed
  File "H5VLcallback.c", line 3571, in H5VL__file_create
    file create failed
  File "H5VLnative_file.c", line 94, in H5VL__native_file_create
    unable to create file
  File "H5Fint.c", line 1833, in H5F_open
    unable to truncate a file which is already open

End of HDF5 error back trace

Unable to open/create file 'three_sources.h5'
```

I suggest that files stay open only as long they are used, and suggest to refactor the code the following: 

* The `h5f` trait becomes a `Property,` and the getter method is a context manager handling the opening and closing of the file. Especially the latter is important when the generator finishes 
* The `data` trait becomes a `Property` which has a `set` method (assign an array of data instead of reading from file) and a `get` method. The latter returns a copy of the time data. In addition, the `get_data_by_reference` only yields the reference to the data (this is then used inside the result generator).
* use `pathlib` instead of `os.path` module 
* make some internally used methods private (`_load_data`, `_load_metadata`)

### DIFF
--- a/acoular/sources.py
+++ b/acoular/sources.py
@@ -207,9 +207,9 @@ class TimeSamples( SamplesGenerator ):
                 for nodename, nodedata in file.get_child_nodes('/metadata'):
                     self.metadata[nodename] = nodedata
 
-    def _set_data(self, value):
+    def _set_data(self, data):
         self.name = ''
-        self._data = value
+        self._data = data
         self._initialize_timedata()
 
     def _get_data_by_reference(self, file):
--- a/acoular/sources.py
+++ b/acoular/sources.py
@@ -180,11 +180,13 @@ class TimeSamples( SamplesGenerator ):
             self._initialize_timedata()
             self._load_metadata()
         else:
+            self.metadata = {}
             self.numsamples = 0
             self.numchannels = 0
             self.sample_freq = 1.0
-            msg = f"No such file: {self.name}"
-            raise OSError(msg)
+            if self.name != '':
+                msg = f"No such file: {self.name}"
+                raise OSError(msg)
 
     def _initialize_timedata( self ):
         """Initializes the attributes `numchannels`, `numsamples` and `sample_freq` (Only for internal use)."""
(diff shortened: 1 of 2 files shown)


esarradj:
As far as specifically the problem with running acoular_demo twice is considered, I would proposed not to write .h5 file at all and use the output of `SourceMixer` directly as input instead of `TimeSamples` via the .h5 file.

Further, I am a bit concerned how to make sure that we can always read from the file when necessary. For instance, when the generator is not exhausted and then restarted or we have two separate instances of the generator running from the same object.

adku1173:
> As far as specifically the problem with running acoular_demo twice is considered, I would propose not to write .h5 file at all and use the output of `SourceMixer` directly as input instead of `TimeSamples` via the .h5 file.
> 

That might be an option. 

> Further, I am a bit concerned how to make sure that we can always read from the file when necessary. For instance, when the generator is not exhausted and then restarted or we have two separate instances of the generator running from the same object.

I've tested it without running into problems. Having multiple readers is a common case and according to this [conversation](https://forum.hdfgroup.org/t/concurrent-access/968/2), opening the same file multiple times should not be an issue for the HDF5 core library. 

From the h5py docs: _Read-only parallel access to HDF5 files works with no special preparation: each process should open the file independently and read data normally (avoid opening the file and then forking)._ 

At first glance, I don't see any problem here as long as we don't think about process parallelism (https://groups.google.com/g/h5py/c/bJVtWdFtZQM?pli=1). Unfortunately the H5py and PyTables docs are not too informative...

esarradj:
#262 removes the problem with calling `acoular_demo.run()` twice

Issue #226, opened 2024-04-22
Unused arguments in some functions and methods


adku1173:
There exist some methods and functions with unused arguments (For most of these methods, this was intended).
Here we need to decide if it is really necessary to keep unused arguments (e.g. in fastFuncs.py)

Output: 
```
acoular/environments.py:244:18: ARG002 Unused method argument: `xx`
acoular/environments.py:623:16: ARG001 Unused function argument: `t`
acoular/fastFuncs.py:523:33: ARG001 Unused function argument: `distGridToArrayCenter`
acoular/fastFuncs.py:573:38: ARG001 Unused function argument: `distGridToArrayCenter`
acoular/h5files.py:45:33: ARG002 Unused method argument: `group`
acoular/nidaqimport.py:159:35: ARG002 Unused method argument: `dispatch`
acoular/sdinput.py:65:33: ARG002 Unused method argument: `change`
acoular/sources.py:1254:36: ARG002 Unused method argument: `event`
acoular/tfastfuncs.py:88:18: ARG001 Unused function argument: `r0`
acoular/tfastfuncs.py:129:19: ARG001 Unused function argument: `r0`
acoular/tools.py:210:23: ARG001 Unused function argument: `derivativearccos`
acoular/tprocess.py:465:30: ARG002 Unused method argument: `x0`
acoular/tprocess.py:752:48: ARG002 Unused method argument: `method`
acoular/tprocess.py:752:56: ARG002 Unused method argument: `array_dimension`
acoular/tprocess.py:752:73: ARG002 Unused method argument: `interp_at_zero`
acoular/tprocess.py:899:62: ARG002 Unused method argument: `Q`
acoular/tprocess.py:899:67: ARG002 Unused method argument: `interp_at_zero`
acoular/tprocess.py:1235:36: ARG002 Unused method argument: `event`
```

adku1173:
The `SpatialInterpolator` class includes the public attributes `interp_at_zero` and `Q`. 

https://github.com/acoular/acoular/blob/d06d58c6f8101ff9ff186905e43802cc19f5b058/acoular/tprocess.py#L720-L721

At the same time, the method `_result_core_func` has arguments `interp_at_zero` and `Q`, which are never used inside the function code (only the class attribute is used via `self.interp_at_zero`:

https://github.com/acoular/acoular/blob/d06d58c6f8101ff9ff186905e43802cc19f5b058/acoular/tprocess.py#L913-L1134

[...]

PR #244, opened 2024-05-08
Remove outdated submodules

esarradj:
This PR removes two outdated submodules. `fileimport` used to translate foreign file formats to acoular *.h5, but was not updated for a very long time. It is untested, unsupported and not documented. The file formats may have changed. `nidaqimport` handles the NIDAQ api to read data from NI hardware. This Python2 submodule was not updated and never tested with Python3.

References to both modules are also removed. Addresses part of #226 .


artpelling (approved):
Looks good as far as I can tell!

PR #249, opened 2024-05-18
Adds new UMA-16 xml file

### LINKED ISSUE #63: Wrong coordinates in supplied UMA16 mic layout file
Hi and thank you for acoular

I did some simple tests and I suspect the UMA16 mic layout file `minidsp_uma16.xml`, 
which comes with acoular has the wrong coordinates or the wrong coordinate system orientation.

I am assuming that the sound pressure map is supposed to map sources when looking from the microphone array onto the sources like a camera (A below)? 
However, wher testing acoular with UMA16, I get a result, which looks like a map of the pressure of the sources looking into the array (B below).

(A) Me -> Array -> Src
(B)            Array <- Src <- Me

This is how I tested the UMA16 array.

The UMA16 looks like a 3x3 tic-tac-toe game with a mic placed on each crossing of lines.
```
 - - -
| | | |
 - - -
| | | |
 - - -
| | | |
 - - -
```
For my test I placed a small speaker at the top-right corner of the UMA 16 array 
(looking direction as in (A): Me -> Array -> Src), and I drove it with a 3.5 kHz sine wave. 
When I run accoular (code below), the source appears as an image in the top-left corner.
```
 Test         Estimate
 - - -         - - - 
| | |O|       |O| | |
 - - -         - - - 
| | | |       | | | |
 - - -         - - - 
| | | |       | | | |
 - - -         - - - 
```

I can only explain this result, if for some reason, acoular implements the mapping as in (B): Array <- Src <- Me.
Is this how it is supposed to work?

To clarify things, I looked into the `minidsp_uma16.xml`. 
Assuming the coordinates in the xml-file are in an XY coordinate system, 
where X(left->right); Y(bottom->up) the arrangement of the microphones is as follows: 
```
8-7-10-9
| | |  |
6-5-12-11
| | |  |
4-3-14-13
| | |  |
2-1-16-15
```
On the physical array this corresponds to looking into the array as in (B) Array <- Src <- Me.
When I use the array, however, I rotate the array 180 deg around the vertical axis 
to point it to the source. This corresponds to configuraton (A): Me -> Array -> Src.
When I look from behind the array, I see the microphones arranged like this:
```
9--10-7-8
|  |  | |
11-12-5-6
|  |  | |
13-14-3-4
|  |  | |
15-16-1-2

```
Going back from the physical world to the world of arbitrary coordinate system choices,
this physical situation can still be described by the original XY coordinate system, 
considering it being rotated 180 deg around the vertical Y-axis. 
The coordinates system now has an inverse x-axis: X(left<-right); Y(bottom->up).

However, in actual test, this produced the flipped image of the sound source as described above.
Flipping the microphone position matrix left-to-right solves the problem.

So, this makes me think that the UMA16 mic layout file `minidsp_uma16.xml` has the mic coordinates mistaken.

Would you please clarify this point?

My code is below:

```
# Note: filenames and paths may need re-adjustment

temp_h5_datafile = 'uma16_test_quadrant1.h5'
micgeofile = 'minidsp_uma16.xml'
mg = acoular.MicGeom(from_file=micgeofile)
ts = acoular.TimeSamples(name=temp_h5_datafile) 
ps = acoular.PowerSpectra(time_data=ts, block_size=128, window='Hamming')
rg = acoular.RectGrid(x_min=-0.2, x_max=0.2, y_min=-0.2, y_max=0.2, z=0.12,
                      increment=0.01)
st = acoular.SteeringVector(grid=rg, mics=mg)
bb = acoular.BeamformerBase(freq_data=ps, steer=st)
pm = bb.synthetic(3500, 3)
Lm = acoular.L_p(pm)

# Generate the figure
plt.imshow(Lm.T, origin='lower', vmin=Lm.max()-10, extent=rg.extend(),
       interpolation='bicubic')
plt.colorbar()
plt.figure(1)

plt.plot(mg.mpos[0], mg.mpos[1], 'o')
# for i in range(0, mg.mpos[0].size):
for i in range(0,5):  # label only the first 5 of the points
    x = mg.mpos[0][i]
    y = mg.mpos[1][i]
    plt.text(x+0.001, y+0.005, i+1, fontsize=9, color='yellow')
plt.axis('equal')
plt.show()

# Close all open .h5 files if any and remove the temporary data file
tables.file._open_files.close_all()

```

adku1173:
* changes "minidsp_uma16.xml" file to "minidsp_uma16_mirrored.xml"
* adds new corrected "minidsp_uma16.xml" file
* adds new `export_mpos` method to `MicGeom` class
* closes #63


acoular/xml/minidsp_uma16.xml
gherold, line 1: I would suggest to not use the exact same name of the original (mirrored) file (to avoid "silent" errors in scripts that use this). Better rename this to `uma16.xml` or `minidsp_uma-16.xml`. The original filename should just not exist anymore.

acoular/xml/minidsp_uma16_mirrored.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?><MicArray name="uma16">
+   <!-- Contains the sensor positions for the UMA-16 MEMS microphone array from miniDSP. -->
+   <!-- The sensor positions are mirrored along the x-axis (see: GitHub Issue #63). -->
+   <!-- This file was named minidsp_um16.xml file until Acoular version 24.03. -->
```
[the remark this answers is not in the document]
adku1173: fixed now.

acoular/xml/minidsp_uma16.xml
adku1173: new filenames are `minidsp_uma-16.xml` and `minidsp_uma-16_mirrored.xml`

gherold (approved):

PR #252, opened 2024-05-21
Py312

adku1173:
same PR as #251 -> this PR is only to check potential permission problems with forks of the repo

### DIFF
--- a/.github/actions/setup-hatch/action.yml
+++ b/.github/actions/setup-hatch/action.yml
@@ -4,7 +4,7 @@ inputs:
   python-version:
     description: 'The Python version'
     required: false
-    default: '3.11'
+    default: '3.12'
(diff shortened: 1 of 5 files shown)


adku1173:
@artpelling there seem to be a problem with the permissions -> https://github.com/actions/labeler#permissions. The same PR (#251) fails when received from a fork.

artpelling:
It only fails if there is no label which is intended. The issue is that the auto-labeler cannot assign labels to PRs made from forks.
A solution could be to mirror external PRs with a local branch (can be done in CI). @esarradj and I already had a conversation about this.
For the moment, however, this is probably overdone. Main devs can/should just work on branches in the main repo and for external PRs we can just assign a label by hand.

PR #255, opened 2024-05-21
remove plot_example

artpelling:
I think this file is unused currently and we can remove the entire directory.


esarradj:
I think we might need this helper for extending the documentation with all those currently undocumented examples. So the script should be moved to examples not just deleted.

esarradj (approved):

esarradj (approved):

PR #256, opened 2024-05-21
Bugfix GitHub pages deploy

adku1173:
* fixes missing artifacts after download by explicitly specifying download path
* fixes sphinx warnings encountered when docs are build via `hatch -v run docs:build`
* fixes missing autosummary for `acoular.tools`, `acoular.demo` submodules


.ruff.toml
```diff
@@ -61,7 +61,7 @@ fixable = ["ALL"]
 unfixable = []
 
 [lint.pydocstyle]
-convention = "google"  # Accepts: "google", "numpy", or "pep257".
```
artpelling, line 64: I wondered about that one already. 😸

.github/workflows/docs.yml
```diff
@@ -7,11 +7,7 @@ on:
     branches: [ master ] # build docs as additional test on PRs to main
   workflow_dispatch:
     branches: [ '*' ] # allow manual trigger for all branches
-
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
-
```
artpelling, line 14: Why did you remove concurrency?

artpelling (approved):

.github/workflows/docs.yml
adku1173: ups, this was not intended. Should be fixed now!

artpelling (approved):

PR #259, opened 2024-05-23
CI for TestPyPI and PyPI

adku1173:
* uses hatch to build wheel and tarball
* adds CI workflows to build and deploy to TestPyPI and PyPI 
* fixes some typos and problems with links and images in README.md
* successful test publishing on TestPyPI -> https://test.pypi.org/project/acoular/ (I had to temporarily use a different version number, because after the upload the respective version is blocked and cannot be used again)

One thing I wasn't sure about was how we should handle it is what the tarball archive needs to include. Probably, the package could be smaller if we exclude some directories such as the tests directory for example. 

One flaw is that the installation with testpypi does not work out of the box, as some package distributions that Acoular depends on are not on TestPyPI. Therefore I had to check locally if the wheels and tarball are correct.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@
 
 [project]
 name = "acoular"
-version = "0.2"
+version = "24.05"
 description = "Python library for acoustic beamforming"
 requires-python = ">=3.8,<=12"
 authors = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,8 +54,8 @@ full = [
 ]
 
 [project.urls]
-homepage = "http://acoular.org"
-documentation = "http://acoular.org"
+homepage = "https://acoular.org"
+documentation = "https://acoular.org"
 repository = "https://github.com/acoular/acoular"
 
 [build-system]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -54,8 +54,8 @@ full = [
 ]
 
 [project.urls]
-homepage = "https://acoular.org"
```
esarradj, line 57: https is correct IMHO

pyproject.toml
```diff
@@ -101,3 +101,13 @@ coverage = ["test --cov=acoular"]
 platform.linux.pre-install-commands = ['cat /proc/cpuinfo']
 platform.macos.pre-install-commands = ['sysctl -a machdep.cpu']
 platform.windows.pre-install-commands = ['systeminfo']
+
+
+[tool.hatch.build.targets.sdist]
+exclude = [
```
esarradj, line 107: Only what is needed to build the package should be included in the sdist. I think that is just ./acoular, pyproject.toml and some other files in the project root (approx 150 kB). The last sdist on pypi includes everything including test data and is stunning 7 MB large.

adku1173: We handled it differently across various releases (24.03, 23.11, 23.06 -> approx. 7 MB, 22.03, 21.05, ... < 1 MB).

pyproject.toml
adku1173: fixed

pyproject.toml
adku1173: now includes only the source directory

esarradj (approved):

PR #261, opened 2024-05-24
use hatch for publishing package

adku1173:
* use `hatch publish` and GitHub secret Token to publish on TestPyPI and PyPI


.github/workflows/testpypi.yml
```diff
@@ -21,27 +21,5 @@ jobs:
...
-      uses: actions/download-artifact@v4
-      with:
-        name: python-package-distributions
-        path: dist/
     - name: Publish distribution 📦 to TestPyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        repository-url: https://test.pypi.org/legacy/          
\ No newline at end of file
+      run: hatch publish -r test -u __token__ -a ${{ secrets.HATCH_PUBLISH_TESTPYPI_API_TOKEN }}
```
artpelling, line 25: where is the test repo defined? shouldn't there be an entry in `pyproject.toml`?

adku1173: It has nothing to do with the pyproject.toml but with the repository (TestPyPI in this case)

gherold (approved):

PR #262, opened 2024-05-24
Updates demo with ASCII map and gets rid of writing file

esarradj:
The demo will no longer write the synthesized time data to file.
The demo will print an ASCII if matplotlib is not available. 
Docstring error is corrected.


esarradj:
It turns out that  'three_sources.h5' are needed for the getting started documentation. Maybe this could be changed in the future. For now I leave it.

artpelling (approved):

PR #265, opened 2024-06-02
Adds flow establishment length traits

### LINKED ISSUE #234: Assure flow and plane unit vectors in SlotJet class
This issue arises from the following to-do written by @esarradj :
https://github.com/acoular/acoular/blob/1ea64301a0413c4f58d837703b5c42a3e57e2540/acoular/environments.py#L339-L340

esarradj:
Adds new traits to `OpenJet` and `SlotJet` to adjust the length of flow establishment zone if needed.
Also closes #234 - it turned out that there is no need of unit vectors with the present code.

### DIFF
--- a/acoular/environments.py
+++ b/acoular/environments.py
@@ -303,9 +303,12 @@ class SlotJet(FlowField):
     #: Width of the slot, defaults to 0.2 .
     B = Float(0.2, desc='nozzle diameter')
 
+    #: Nondimensional length of zone of flow establishment (jet core length), defaults to 5.2
+    l = Float(5.2, desc='flow establishment length')
+
     # internal identifier
     digest = Property(
-        depends_on=['v0', 'origin', 'flow', 'plane', 'B'],
+        depends_on=['v0', 'origin', 'flow', 'plane', 'B', 'l'],
     )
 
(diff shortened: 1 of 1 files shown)


adku1173 (changes requested):
If there is no need for unit vectors, this PR should remove the following TODO note
https://github.com/acoular/acoular/blob/d76d33910f24269280a5e79b2a3c1a5491ed0781/acoular/environments.py#L336-L337

acoular/environments.py
```diff
@@ -382,9 +385,12 @@ class OpenJet(FlowField):
     #: Diameter of the nozzle, defaults to 0.2 .
     D = Float(0.2, desc='nozzle diameter')
 
+    #: Nondimensional length of zone of flow establishment (jet core length), defaults to 6.2
+    l = Float(6.2, desc='flow establishment length')
```
adku1173, line 389: What is the reason for the different default values between `SlotJet` and `OpenJet`

esarradj: This is indeed different for slot jet and round jet according to Albertson1950.

adku1173 (approved):

PR #270, opened 2024-06-11
make SoundDeviceSamplesGenerator always importable

### LINKED ISSUE #211: Improve handling of optional classes
There exist some optional imports in the [init.py](acoular/__init__.py#L19-27), which are problematic for the following reason:

* bare except is bad code style and a linting error (E722)
* classes are simply missing if the required packages are not installed (I think we should at least raise an ImportError when the class is imported by a user)

adku1173:
closes #211

### DIFF
--- a/acoular/__init__.py
+++ b/acoular/__init__.py
@@ -6,14 +6,9 @@
 
 import os
 
-from .configuration import config
-from .version import __author__, __date__, __version__
-
-if config.have_sounddevice:
-    from .sdinput import SoundDeviceSamplesGenerator
-
 from . import demo, tools
 from .calib import Calib
+from .configuration import config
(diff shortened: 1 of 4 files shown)


acoular/__init__.py
```diff
@@ -126,3 +122,4 @@
     WriteWAV,
 )
 from .trajectory import Trajectory
+from .version import __author__, __date__, __version__
```
artpelling, line 125: why is this needed?

artpelling (approved):

acoular/__init__.py
adku1173: It is needed to call `acoular.__version__` etc.

PR #280, opened 2024-06-27
Implement a lazy result array for BeamformerBase derived classes

### LINKED ISSUE #151: Rethink CSM / PowerSpectra valid frequency indexing (again)
Current PowerSpectra.freq_range adds easier access than ind_low/ind_high to limit calculated frequencies, but still is not very user-friendly if user
- wants to calculate only one discrete frequency band (has to find next frequency to add as upper limit)
- wants to calculate a list of discrete frequencies, omitting some in-between (not currently possible)
- wants to calculate a distinct band width around frequencies, e.g. 3rd-octave (freq limits have to be calculated "by hand")

The cleanest solution would be to let the Beamformer decide which frequencies will be calculated (depending on what is asked).

### LINKED ISSUE #175: New Beamformer-class function "condense" for replacing/enhancing synthetic+integrate?
Suggestion to add a new function in the BeamformerBase function that comprises the functionality of both ".synthetic" and ".integrate". Purpose is to compress the result data and extract the actually interesting info (e.g. third-octave bands in defined sectors).

Most of the arguments passed to the function should be optional, so that it can be used more or less in the same way as .synthetic or .integrate alone. This would also make using the corresponding standalone functions after first applying .synthetic or .integrate unnecessary.

Important differences to the old functions are
- no reshaping to grid anymore (should be done somewhere else)
- ability to handle multiple requested frequency bands and sectors (but also still single ones)

The implementation of this could include a deprecation warning for .synthetic and .integrate

Any thoughts or naming ideas? (condense/compress/extract ... )

### LINKED ISSUE #266: Confusion about digest and ext_digest in BeamformerBase
I am slightly confused that we have two digests in the `BeamformerBase` class: 

https://github.com/acoular/acoular/blob/a12e7e9d93325ed53f5f273d41cf9548e436dfd2/acoular/fbeamform.py#L393-L400

Why is this necessary @esarradj @gherold ? I guess, I just missed something.

Can't we simply include the dependence on `freq_data.ind_low` and `freq_data.ind_high` in the normal digest? 
In addition:  why not move the `ind_low` and `ind_high` attributes from `PowerSpectra` to the Beamformer classes since `ind_low` and `ind_high` are irrelevant for the power spectra calculation and are only used for calculating the beamforming result in the respective frequency range? (I know we should keep backward compatibility, but it doesn't seem logical to me)

esarradj:
This addresses several issues by reworking some basic mechanics of `BeamformerBase`. In `BeamformerBase` and all derived classes, the result is computed per frequency. As the computations are quite expensive especially for advanced methods, it is inefficient to compute the result for any frequency that is not needed or required in the end. So far, this was controlled by `ind_low` and `ind_high` attributes of the `freq_data` attribute to `BeamformerBase` which lead to problems.  This closes #151 
by introducing a mechanism which computes only for those frequencies that are actually needed when then `result` attribute is queried using the index operator []:

- new private helper class `LazyResultArray` which controls the computation
- `_calc` method replaces 'calc' method
- `BeamformerBase` gets non-public, non-trait attributes to avoid passing them around unnecessarily: `_ac`, `_fr`, `_f` 
- `integrate` method gets new functionality and is now able to integrate _and_ deliver the result as n-th octave spectrum or for just a limited frequency range of interest; this closes #175
- `ext_digest` is no longer necessary, closes #266

While this works for all other `BeamformerBase` derived classes, `BeamformerSODIX` received special treatment:

- `_calc` also computes all frequencies below those actually requested as those results are needed as an initial value for subsequent frequencies
- `sodix_result` is now replaced by `result` as in all other classes
- special versions of `_get_file_cache` and `synthetic` are no longer needed

As a side effect, this pull request leads to a considerable speed-up for the beamformer tests, which need now only 25% of the time to finish.

### DIFF
--- a/acoular/fbeamform.py
+++ b/acoular/fbeamform.py
@@ -1309,56 +1309,46 @@ class BeamformerDamas(BeamformerBase):
         self.r_diag = self.beamformer.r_diag
         self.steer = self.beamformer.steer
 
-    def calc(self, ac, fr):
-        """Calculates the DAMAS result for the frequencies defined by :attr:`freq_data`.
+    def _calc(self, ind):
+        """Calculates the result for the frequencies defined by :attr:`freq_data`.
 
         This is an internal helper function that is automatically called when
-        accessing the beamformer's :attr:`~BeamformerBase.result` or calling
-        its :meth:`~BeamformerBase.synthetic` method.
-        A Gauss-Seidel algorithm implemented in C is used for computing the result.
+        accessing the beamformer's :attr:`result` or calling
+        its :meth:`synthetic` method.
 
         Parameters
         ----------
-        ac : array of floats
-            This array of dimension ([number of frequencies]x[number of gridpoints])
-            is used as call-by-reference parameter and contains the calculated
-            value after calling this method.
-        fr : array of booleans
-            The entries of this [number of frequencies]-sized array are either
-            'True' (if the result for this frequency has already been calculated)
-            or 'False' (for the frequencies where the result has yet to be calculated).
-            After the calculation at a certain frequency the value will be set
-            to 'True'
+        ind : array of int
+            This array contains all frequency indices for which (re)calculation is
+            to be performed
 
         Returns
         -------
-        This method only returns values through the *ac* and *fr* parameters
+        This method only returns values through :attr:`_ac` and :attr:`_fr`
 
         """
-        f = self.freq_data.fftfreq()
+        f = self._f
         normfactor = self.sig_loss_norm()
-
         p = PointSpreadFunction(steer=self.steer, calcmode=self.calcmode, precision=self.psf_precision)
         param_steer_type, steer_vector = self._beamformer_params()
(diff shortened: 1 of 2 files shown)


acoular/fbeamform.py
```diff
@@ -256,6 +258,34 @@ def steer_vector(self, f, ind=None):
         return func(self.transfer(f, ind))
 
 
+class LazyBfResult:
+    """Manages lazy per-frequency evaluation."""
+
+    # Internal helper class which works togther with BeamformerBase to provide
```
gherold, line 264: typo: "together"

acoular/fbeamform.py
gherold, line 1: Line 640ff (` if ind not in indices:
                    warn`...): Is there still a case where this Warning could be triggered? If not, either add a comment that this is redundant or remove conditional completely

gherold (commented):
Small changes

gherold (approved):

PR #285, opened 2024-07-05
Streaming improvements

jake-is-ESD-protected:
I noticed that acoular's compatibility with physical hardware is underdeveloped, so I propose this PR which includes the following features:

- `SoundDeviceSamplesGenerator` can now take the argument `data_type`, which is the audio resolution/format for the requested audio stream from a real device. This works, because `sd` supports it. The formats should be passed as strings. The supported values are `"float32"`, `"int32"`, `"int16"`, `"int8`, `"uint8_t"`.
- `PowerSpectra` was modified to include yieldable output for unlimited data streams. `num_blocks` now has a shadow trait which can be set by the user if the given source is a stream. `num_blocks` then represents how many CSMs should be averaged per retrieval. An internal trait called `_should_yield` is set to `True` if the spectra's source has a `numsamples` of `-1`, which means "keep streaming"

Tests all run. For some reason, the `.gitignore` did not ignore `.venv`, so I added that.


acoular/sdinput.py
```diff
@@ -49,6 +49,11 @@ def __init__(self, *args, **kwargs):
     #: Sampling frequency of the signal, changes with sinusdevices
     sample_freq = Property(desc='sampling frequency')
 
+    _sample_freq = Int(0)
+
+    #: Datatype (resolution) of the signal, used as `dtype` in a sd `Stream` object
+    data_type = String('float32', desc='data type (resolution) of signal')
```
esarradj, line 55: This should be an Enum of possible values and should also allow certain numpy.dtype value (see sd docs)

acoular/sdinput.py
```diff
@@ -49,6 +49,11 @@ def __init__(self, *args, **kwargs):
     #: Sampling frequency of the signal, changes with sinusdevices
     sample_freq = Property(desc='sampling frequency')
 
+    _sample_freq = Int(0)
```
esarradj, line 52: should be Float

acoular/sdinput.py
```diff
@@ -71,7 +76,12 @@ def _get_numchannels(self, event):  # noqa ARG002
         self.numchannels = min(self.numchannels, sd.query_devices(self.device)['max_input_channels'])
 
     def _get_sample_freq(self):
+        if self._sample_freq:
```
esarradj, line 79: if _sample_freq is a float, this will not work

acoular/spectra.py
```diff
@@ -384,17 +399,28 @@ def calc_csm(self):
...
         for data in self.get_source_data():
+            i+=1
             ft = fft.rfft(data * wind, None, 0).astype(self.precision)
             calcCSM(csm_upper, ft)  # only upper triangular part of matrix is calculated (for speed reasons)
+            if i == self.num_blocks and self._should_yield:
+                i = 0
+                return self._yield_csm(csm_upper, weight)
+        return self._full_csm(csm_upper, weight)
+    
+    def _yield_csm(self, csm_upper, weight):
```
esarradj, line 420: This gives a 'one time' generator which is not compatible to anything located downstream, therefore this change should be reverted.

acoular/spectra.py
```diff
@@ -294,8 +300,17 @@ class PowerSpectra(BaseSpectra):
 
     @property_depends_on('_source.numsamples, block_size, overlap')
     def _get_num_blocks(self):
+        if self._num_blocks != 0: # The source is a real time stream
```
esarradj, line 303: This could also be used for non real time data.

acoular/spectra.py
```diff
@@ -294,8 +300,17 @@ class PowerSpectra(BaseSpectra):
...
+            return self._num_blocks 
         return self.overlap_ * self._source.numsamples / self.block_size - self.overlap_ + 1
 
+    def _set_num_blocks(self, value):
```
esarradj, line 307: This should simply set the shadow trait, because setting num_blocks could also be of use for non streaming data. However, it should never be larger than `self.overlap_ * self._source.numsamples / self.block_size - self.overlap_ + 1`.

acoular/spectra.py
```diff
@@ -384,17 +399,28 @@ def calc_csm(self):
...
-        # print "num blocks", self.num_blocks
+        # print( "num blocks", self.num_blocks)
         # for backward compatibility
         if self.calib and self.calib.num_mics > 0:
             if self.calib.num_mics == t.numchannels:
                 wind = wind * self.calib.data[newaxis, :]
             else:
                 raise ValueError('Calibration data not compatible: %i, %i' % (self.calib.num_mics, t.numchannels))
         # get time data blockwise
+        i = 0
```
esarradj, line 410: better to use `enumerate` in the for statement

esarradj (changes requested):
The `yield` functionality does not really work here and should be removed from `PowerSpectra`. Instead a class which is derived from both `BaseSpectra` and `TimeInOut` could be used (see `FFTSpectra`)

esarradj:
Linting check fails, please use `hatch fmt` to solve this.

jake-is-ESD-protected:
I will create a new class based on `BaseSpectra` and `TimeInOut` which has a yieldable `result()` method. The number of blocks will also be settable.

jake-is-ESD-protected:
> Linting check fails, please use `hatch fmt` to solve this.

Please add this information to the `CONTRIBUTING` file as well!

acoular/sdinput.py
jake-is-ESD-protected: Resolved in https://github.com/acoular/acoular/pull/285/commits/ec664a66f6f7ac2200f26f2d35283bc9418d92e0 as proposed

acoular/sdinput.py
jake-is-ESD-protected: Resolved in https://github.com/acoular/acoular/pull/285/commits/ec664a66f6f7ac2200f26f2d35283bc9418d92e0 as proposed

acoular/spectra.py
jake-is-ESD-protected: Reverted

acoular/spectra.py
jake-is-ESD-protected: Resolved in https://github.com/acoular/acoular/pull/285/commits/e156e01325b09ece2aa8ac2ba9572accd96e24af as proposed

acoular/spectra.py
jake-is-ESD-protected: Added a check for this condition https://github.com/acoular/acoular/pull/285/commits/e156e01325b09ece2aa8ac2ba9572accd96e24af

acoular/spectra.py
jake-is-ESD-protected: This is a bit tricky. Using `enumerate` will cause the enumeration to be infinite, which means an overflow at some point. If I then use a `%` or rather a `divmod` operation, the distance between the last multiple of `num_blocks` before and the first after the overflow will not be equal to `num_blocks`, introducing a **latency bug** for long data streams.

Example with a `uint8_t` counter:
```python
num_blocks = 10
for i, data in enumerate(gen):  # assume that i is 8 bit
     q, r = divmod(i, num_blocks)
     process(data)
     if r == 0:
        postprocess()
```
The last time `postprocess()` is called before overflow is at `i = 250`. The next time round it will be called at `i = 10`, which means that between the iterations across the overflow, `process(data)` was called 16 times instead of 10 times. This bug is rare, but it is one nonetheless.

esarradj: I do not agree. `i` as returned by the `enumerate` generator is a Python int. In Python 3 this is unbounded (no upper limit). Even in a hypothetical 'lite' Python implementation this would be something like the processor word size (nowadays at least 16 bits I assume) - this was the way it was in Python 2.

esarradj (approved):

esarradj (changes requested):
consider `num_blocks` or `_num_blocks` in `calc_csm`

jake-is-ESD-protected:
I see that there has been a misunderstanding regarding the implementation of `num_blocks`. The new commit [6af95f6](https://github.com/acoular/acoular/pull/285/commits/6af95f678bd75212994065fd8994607b3258eca2) fixes this by splitting the total amount of possible blocks (depending on `num_samples` and `block_size`) into the requested amount of CSMs.

### Example
A power spectra with a block size of 128, block number of 4 and a sample source with 1024 samples will now return CSMs of shape `(numfreq, 2, numchannels, numchannels)`, where axis 1 are the CSMs consisting only of the number of requested blocks per CSM (4 blocks of 128 samples occur 2 times in 1024 samples)

### Further Questions
Right now, the average still happens over all blocks. I do not know if this is correct, because I could also imagine that the time average now has to take place over each set of blocks instead of all blocks. Can you confirm this? Additionally, this approach is vulnerable towards a `num_blocks` value that is not an even integer division of the total amount of blocks in the signal. Is it viable from the user to request a `num_blocks` of 6 when the total amount of blocks is 8? This would result in an incomplete set of blocks at the end. Should this be caught? I think this is only needed in case that the previous question resolves to "time average per set of blocks"

acoular/spectra.py
jake-is-ESD-protected: The proposed improvements for the `PowerSpectra` class are based on `range` instead ([6af95f6](https://github.com/acoular/acoular/pull/285/commits/6af95f678bd75212994065fd8994607b3258eca2)). However, I will use `enumerate()` in the not yet pushed `PowerSpectraInOut` class.

acoular/spectra.py
```diff
@@ -9,6 +9,7 @@
     BaseSpectra
     FFTSpectra
     PowerSpectra
+    PowerSpectraInOut
```
esarradj, line 12: This class does not (yet) exist in Acoular.

acoular/spectra.py
```diff
@@ -392,13 +409,23 @@ def calc_csm(self):
...
+                calcCSM(csm_upper, ft)  # only upper triangular part of matrix is calculated (for speed reasons)
+            # create the full csm matrix via transposing and complex conj.
+            csm_lower = csm_upper.conj().transpose(0, 2, 1)
+            [fill_diagonal(csm_lower[cntFreq, :, :], 0) for cntFreq in range(csm_lower.shape[0])]
+            csm_block = csm_lower + csm_upper
+            if n_groups > 1:
+                csm_block = csm_block[:, newaxis, :, :]
+                csm = concatenate([csm, csm_block], axis=1)
+            else:
+                csm = csm_block
```
esarradj, line 428: Two problems here:

1. breaks API by possibly returning a four-dimensional array. What we need is just one CSM per frequency, and the num_blocks just helps to limit the number of blocks (one time) in case we have an infinite stream as the input. Full streaming cannot be done anyway with the `BeamformerBase` derived classes.
2. the code is very tortuous and difficult to check
```suggestion
        for num, data in enumerate(self.get_source_data()):
            if num > self.num_blocks:
                break
            ft = fft.rfft(data * wind, None, 0).astype(self.precision)
            calcCSM(csm_upper, ft)  # only upper triangular part of matrix is calculated (for speed reasons)
        # create the full csm matrix via transposing and complex conj.
        csm_lower = csm_upper.conj().transpose(0, 2, 1)
        [fill_diagonal(csm_lower[cntFreq, :, :], 0) for cntFreq in range(csm_lower.shape[0])]
        csm = csm_lower + csm_upper
        # in case there where less blocks available
        if num < self.num_blocks:
            return csm * (2.0 / self.block_size / weight / num)
```

esarradj (changes requested):

acoular/sdinput.py
```diff
@@ -49,6 +50,11 @@ def __init__(self, *args, **kwargs):
     #: Sampling frequency of the signal, changes with sinusdevices
     sample_freq = Property(desc='sampling frequency')
 
+    _sample_freq = Float(0)
+
+    #: Datatype (resolution) of the signal, used as `dtype` in a sd `Stream` object
+    data_type = Enum([float32, float16, int32, int16, int8, uint8], desc='data type (resolution) of signal')
```
adku1173, line 56: To be consistent with other classes in Acoular, this trait should be named `precision` and the type should be `str`. The sounddevice `InputStream` object accepts a string for the `dtype` argument 

Thus: `precision = Trait('float32', 'float16',...)`

acoular/spectra.py
```diff
@@ -253,6 +256,9 @@ class PowerSpectra(BaseSpectra):
     #: (set from block_size and overlap).
     num_blocks = Property(desc='overall number of FFT blocks')
 
+    # Is set to a user value if the given SamplesGenerator streams non-finite data
```
adku1173, line 259: There is no need to adjust the number of blocks that the `PowerSpectra` object should average. Instead, the classes `MaskedTimeSamples`  and `MaskedTimeInOut` exists, which can be used to set the evaluation time interval. Thus, I suggest to revert this change.

adku1173 (changes requested):
I recommend (first) concentrating on the changes to the `SoundDeviceSamplesGenerator` class in this PR. There is another PR (#294)  which deals with blockwise frequency domain processing and will handle the generation of the CSM.

acoular/spectra.py
```diff
@@ -294,8 +300,19 @@ class PowerSpectra(BaseSpectra):
 
     @property_depends_on('_source.numsamples, block_size, overlap')
     def _get_num_blocks(self):
+        if self._num_blocks != 0:
+            return self._num_blocks
         return self.overlap_ * self._source.numsamples / self.block_size - self.overlap_ + 1
 
+    @property_depends_on('_source.numsamples, block_size, overlap')
```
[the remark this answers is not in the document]
jake-is-ESD-protected: Reverted: [8d41784](https://github.com/acoular/acoular/pull/285/commits/8d417847fb9a19be7377bb2cc2756182359a5fdd)

acoular/spectra.py
```diff
@@ -294,8 +300,19 @@ class PowerSpectra(BaseSpectra):
...
+    def _set_num_blocks(self, b):
+        max_nb = self.overlap_ * self._source.numsamples / self.block_size - self.overlap_ + 1
```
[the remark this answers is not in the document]
jake-is-ESD-protected: Reverted: [8d41784](https://github.com/acoular/acoular/pull/285/commits/8d417847fb9a19be7377bb2cc2756182359a5fdd)

acoular/spectra.py
jake-is-ESD-protected: Reverted: [8d41784](https://github.com/acoular/acoular/pull/285/commits/8d417847fb9a19be7377bb2cc2756182359a5fdd)

On a side-note: I thought that an extra dimension was specifically requested by @adku1173, but that must have been a misunderstanding. According to 

> There is no need to adjust the number of blocks that the `PowerSpectra` object should average. Instead, the classes `MaskedTimeSamples` and `MaskedTimeInOut` exists, which can be used to set the evaluation time interval. Thus, I suggest to revert this change.

this is redundant anyway and changes are being made on a lower level: #294

acoular/sdinput.py
```diff
@@ -49,6 +50,11 @@ def __init__(self, *args, **kwargs):
     #: Sampling frequency of the signal, changes with sinusdevices
     sample_freq = Property(desc='sampling frequency')
 
+    _sample_freq = Float(0)
```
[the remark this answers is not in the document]
jake-is-ESD-protected: [b1794fc](https://github.com/acoular/acoular/pull/285/commits/b1794fccc18731f721f87432d0588085f654633f)

acoular/spectra.py
jake-is-ESD-protected: I did not know this! Reverted with any other changes in `spectra.py`: [f00b1b2](https://github.com/acoular/acoular/pull/285/commits/f00b1b2604ccbfce8e5564335e30c11bd937dede)

jake-is-ESD-protected:
It turns out that this branch does a lot less than originally intended. The actual streaming is being developed here #294.

adku1173 (commented):
PR looks good so far. Please don't forget to finish the last two points on the contribution [checklist](https://www.acoular.org/contributing/checklist.html
):

- [x]  Updated the [What’s new](https://www.acoular.org/news/index.html) section in the [documentation](https://github.com/acoular/acoular/blob/master/docs/source/news/index.rst) explaining the changes you made

- [ ] Append yourself to the author list in the [CITATION.cff](https://github.com/acoular/acoular/blob/master/CITATION.cff) file

adku1173 (approved):

PR #294, opened 2024-07-24
Draft: Blockwise frequency domain processing

adku1173:
I would like to discuss a draft that extends the blockwise processing of frequency domain data.

# Current State

The class `FFTSpectra` transforms time data coming from a `SamplesGenerator` derived class into the frequency domain. This is done **blockwise** with the `result` method.

The class has one bug: it breaks the philosophy of the result generator method since no `num` argument currently exists. 

# Extension

`FFTSpectra.result` now has a `num` argument determining the number of spectra yielded. In contrast to `TimeInOut` derived classes, the output is of shape `(num, numchannels, numfrequencies)` instead of `(num, numchannels)`

Theoretically, since `FFTSpectra` is derived from `TimeInOut`, it can be used as a source for existing classes such as `TimeAverage` or `TimePower` to process frequency domain data. Although this did not work out of the box, extending the functionality was easy for the two classes. However, there will always be some classes that cannot handle sources yielding frequency domain data, or it is at least cumbersome to implement processing in both domains. 

I want to discuss two different solutions, both with pros and cons. (Maybe there are more solutions I have not considered so far.)

## Draft 1 – no explicit freq domain definition

This would mean that `TimeInOut` and `SamplesGenerator` derived classes are not limited to a certain domain. It depends on the class if time data and/or frequency domain data can be handled.

I introduced a `_validate_result_block` method, which can be used to determine the source data domain based on the given block shape. If the class processing the data can't handle the domain, an error is raised. 

 Alternatively, one may consider introducing a (private?) `domain` trait that is delegated from the source and validated by the subsequent class. This would make it possible to detect invalid pipelines before processing starts.

**Pros:**
*  No redundant code 

**Cons**:
* The purpose of a class (freq and/or time domain) can be complicated to recognize. (Requires a thorough documentation)

## Draft 2 - explicit freq domain classes

A clear but also extensive addition would be to introduce a `fprocess.py` module by analogy to the `tprocess.py` module. This module would have a `FreqGenerator` and a ` FreqInOut` class. E.g: 

```python


class FreqGenerator(HasPrivateTraits):
    """Base class for any generating signal processing block in frequency domain."""

    #: Sampling frequency of the signal, defaults to 1.0
    sample_freq = Float(1.0, desc='sampling frequency')

    #: Number of channels
    numchannels = CLong

    #: Number of samples
    numsamples = CLong

    #: Number of samples
    numfreq = CLong

    [...]

class FreqInOut(FreqGenerator):
    """Base class for any frequency domain signal processing block."""

    #: Data source; :class:`~acoular.sources.FreqGenerator` or derived object.
    source = Trait(FreqGenerator)


class FreqPower(FreqInOut):
    """Calculates time-depended power of the signal."""

    #: Data source; :class:`~acoular.fprocess.FreqInOut` or derived object.
    source = Trait(FreqInOut)

    def result(self, num):
        """Python generator that yields the output blockwise.

       Returns
        -------
        Squared output of source.
            Yields blocks of shape (num, numchannels, numfreq).
            The last block may be shorter than num.

        """
        for temp in self.source.result(num):
            yield temp * temp.conjugate()
```

**Pros:**
*  very clear interface
* would enable potentially useful subclasses (`FreqSamples` -> read frequency data from .h5 files)

**Cons**:
* redundant (or even doubled code)
* higher maintenance effort

---

Thoughts?  @gherold, @esarradj, @artpelling, @FrankK20

### DIFF
--- a/acoular/tprocess.py
+++ b/acoular/tprocess.py
@@ -1,7 +1,7 @@
 # ------------------------------------------------------------------------------
 # Copyright (c) Acoular Development Team.
 # ------------------------------------------------------------------------------
-"""Implements processing in the time domain.
+"""Implements blockwise processing in the time domain.
 
 .. autosummary::
     :toctree: generated/
@@ -181,6 +181,32 @@ class TimeInOut(SamplesGenerator):
     # internal identifier
     digest = Property(depends_on=['source.digest'])
 
+    def _validate_result_block(self, block, time_data=True, freq_data=False):
+        """Internal helper function to validate the source result block.
+
+        Frequency domain data blocks are of shape (num, numchannels, numfreqs), while time domain
+        data blocks are of shape (num, numchannels). The number of dimensions indicates
+        the type of data.
+        This method handles the validation of a `data` block yielded by the result
+        :meth:`result` method. It checks the dimensions of the block and raises
+        an error if the block has the wrong dimensions / comes from the wrong domain.
+        """
+        if not time_data and block.ndim == 2:  # probably source yields time domain data
+            msg = (f'{self.source.__class__.__name__} yields data with 2 dimensions. '
+                f'{self.__class__.__name__} does not support time domain data to be processed.'
+                )
+            raise NotImplementedError(msg)
+        if not freq_data and block.ndim == 3:  # probably source yields frequency domain data
+            msg = (f'{self.source.__class__.__name__} yields data with 3 dimensions. '
+                f'{self.__class__.__name__} does not support frequency domain data to be processed.'
+                )
+            raise NotImplementedError(msg)
+        expected_dims = [2,3]
+        if block.ndim not in expected_dims:
+            msg = (f'{self.source.__class__.__name__} yields data with {block.ndim} dimensions, '
+                f'but {self.__class__.__name__} expects on of {expected_dims}.')
+            raise ValueError(msg)
+
     @cached_property
     def _get_digest(self):
         return digest(self)
@@ -292,6 +318,8 @@ class MaskedTimeInOut(TimeInOut):
             i = 0
             fblock = True
             for block in self.source.result(num):
+                if i == 0: # validate first block
+                    self._validate_result_block(block, time_data=True, freq_data=False)
                 bs = block.shape[0]
                 i += bs
                 if fblock and i >= start:  # first block in the chosen interval
@@ -365,7 +393,9 @@ class ChannelMixer(TimeInOut):
         else:
             weights = 1
 
-        for block in self.source.result(num):
+        for i, block in enumerate(self.source.result(num)):
+            if i == 0:
+                self._validate_result_block(block, time_data=True, freq_data=True)
             yield sum(weights * block, 1, keepdims=True)
 
 
@@ -1415,7 +1445,7 @@ class TimePower(TimeInOut):
 
         """
         for temp in self.source.result(num):
-            yield temp * temp
+            yield temp * temp.conjugate()
 
 
 class TimeAverage(TimeInOut):
@@ -1467,10 +1497,15 @@ class TimeAverage(TimeInOut):
         """
         nav = self.naverage
         for temp in self.source.result(num * nav):
-            ns, nc = temp.shape
+            ns, nc = temp.shape[:2]
             nso = int(ns / nav)
+            if temp.ndim ==3:
+                nf = temp.shape[2]
+                oshape = (nso, -1, nc, nf)
+            else:
+                oshape = (nso, -1, nc)
             if nso > 0:
-                yield temp[: nso * nav].reshape((nso, -1, nc)).mean(axis=1)
+                yield temp[: nso * nav].reshape(oshape).mean(axis=1)
 
 
 class TimeCumAverage(TimeInOut):
--- a/acoular/spectra.py
+++ b/acoular/spectra.py
@@ -130,7 +130,7 @@ class BaseSpectra(HasPrivateTraits):
         return None
 
     # generator that yields the time data blocks for every channel (with optional overlap)
-    def get_source_data(self):
+    def _get_source_data(self):
         bs = self.block_size
         temp = empty((2 * bs, self.numchannels))
         pos = bs
@@ -147,10 +147,7 @@ class BaseSpectra(HasPrivateTraits):
 
 
 class FFTSpectra(BaseSpectra, TimeInOut):
-    """Provides the spectra of multichannel time data.
-
-    Returns Spectra per block over a Generator.
-    """
+    """Provides the Fast Fourier Transform (FFT) of multichannel time data."""
 
     # internal identifier
     digest = Property(depends_on=['source.digest', 'precision', 'block_size', 'window', 'overlap'])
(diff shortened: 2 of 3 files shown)


gherold:
I would avoid redundant code if possible (so, leaning more towards the philosophy of Draft 1). Also, I suggest to use the output order `(num, numfreqs, numchannels)`, so that the last dimension always represents the channels This would be more in line with the frequency domain outputs (first frequencies, then channels or gridpoints) and make application of corresponding classes and functions more straight-forward.

I do like the idea of having class-specific traits that provide info about allowed dimensionalities of the `source` and corresponding dimensionalities of the `result`. Currently, the user has to make sure that everything fits. This would still permit creating data processing chains that make no sense, but some obvious mistakes may be prevented before processing starts.

artpelling:
I would prefer Option 2: dedicated implementations of time and frequency classes because it would be much clearer for the user what these classes do. Especially when keeping in mind that we aim to expand our user base, it is important to have a well-written and self explanatory core. It would probably also make compatibility checks much easier to perform.

Regarding redundancy: I am not sure if this is a big issue. `InOut`-derived classes that are domain agnostic could just work for both input domains. Classes where this does not apply would then have two implementations.

That being said, I really dislike the idea of having an extra dimension in the returned results for several reasons: 
1. A DFT is a bijection, so each block in time domain should correspond to a block in frequency domain `</mathematician>` . Jokes aside, as a new user this would be really confusing and I don't see the necessity (when going for dedicated frequency domain implementations).
2. Adding a dimension to use existing processing functionalities originally intended for time domain data seems hacky and will probably lead to a lot of unforseen issues.
3. Given that we intend to make acoular more real-time compatible, I suspect that that return data of shape `(num, numchannels, numfrequencies)` would increase overall latency.

gherold:
@artpelling This sounds actually more like an Option 3, which combines the 2 approaches: Avoid redundancy (i.e. use existing classes for both domains) but allow the dimensions to represent different domains. There would be no need for dedicated classes like a `FreqGenerator` or `FreqPower`. Just use `SamplesGenerator` and `TimePower` (the latter might have to be renamed or "aliased" for better comprehensibility), and allow the SamplesGenerator to contain frequency domain data instead of only time domain. The `...InOut` classes might get an additional trait "`domains`", which contains info about the format of the data. This would not only allow time <-> frequency, but e.g. also spatial domain (=channels) <-> mode domain for the second dimension. Adding transforms aside from fourier (if somehow necessary) would also not be that complicated. 

Each class in the processing pipeline would then check if the `source` domains are compatible. Sounds feasible and clean. Am I forgetting something? The only "downside" I see is that initially "block-wise" calculations were introduced so that you don't run into memory problems when processing large amounts of data. The current (time processing) implementations are done in a way that the block size (`num`) does not matter - the result will always be the same since block-overlapping data is "cached" where necessary. With the new approach we will lose some of that flexibility, but right now I don't see whether this is a big issue. So I'd prefer option 3 :)

gherold:
A better alternative for achieving a similar workflow would be to define new base classes `BaseGenerator` and `BaseInOut`, and then derive corresponding `Time...` and `Freq...` classes from that. This would have the advantages that

1. There is no need for redundant implementations (right now the code for `Freq...` and `Time/Samples...` is more or less identical)
2. If you're sure a class can handle all domains you would just need to check for the `Base...` input.
3. You can still do input restrictions for either `Freq...` or `Time...` or both like suggested above.
4. Extending to further domains would be easier.

@adku1173 I do like the fancy new pipeline operators.

artpelling:
>     5. I added the new operators `<` and `>` to the `TimeInOut` and `FreqInOut` classes so that computational pipelines can be built not only by

I would actually prefer the use of `*`/`__mul__`  since a `...Generator` class has to always be at the beginning of the pipeline, there is only one direction that makes sense. So I think it is good to enforce that direction by always going from left to right with an undirectional operator, i.e.
```python 
ta = ac.TimeSamples() * ac.TimePower() * ac.TimeAverage()
```

We could also think about implementing scalar multiplication this way. I am not sure if this is useful but it shouldnt be too hard to do something like
 ```python
c = 2*np.ones(num_channels)
ta = c * ta
```

File: acoular/tprocess.py [hunk 1]
```diff
@@ -185,6 +185,17 @@ class TimeInOut(SamplesGenerator):
...
+    def __gt__(self, receiver):
+        if isinstance(receiver, (TimeInOut)):
+            receiver.source = self
+            return receiver
+        msg = f'Receiving object {receiver.__class__} must be derived from TimeInOut or FreqInOut.'
+        raise TypeError(msg)
+
+    def __lt__(self, source):
+        self.source = source
+        return self
```
artpelling, line 197:
```suggestion
    return source > self
```

esarradj:
I am a bit late to party it seems, but please allow me to chime in. The preliminary solution so far looks not fully convincing to me. 

What do we want to achieve? I would say, we want to have some processing chain where a possibly infinite stream of information enters (time samples only at present - are there any other possibilities?). Between the blocks in this chain we do have data from possibly different domains (time, frequency, modal and any combinations of it). So a `...InOut` derived class could map any of them to any of them, including such things as time-frequency domain (see e.g. `Filterbank` class). Most classes are useful for just one case, others may be useful for multiple or all cases (e.g. `TimeAverage`). Because there are so many possible options, just having `FreqInOut` does not really solve this situation unless we have also `TimeInFreqOut`, `FreqInTimeOut` etc.

The present implementation **in principle already addresses all cases** except where one does not want to work along a possibly infinite time axis. However, can you imagine a method which uses e.g. an infinite frequency axis to work along? For me, the restriction to time seems to be no problem, as this includes e.g. these cases:

- samples of sound pressure (one per sampling period)
- complex-valued Fourier spectra (one spectrum per sampling period)
- real-valued Power spectra (one spectrum per sampling period)
- modal coefficients (one set per sampling period)
- modal-frequency 'spectra' (one per sampling period)

In the extreme case, the sampling period could also be very long ('infinite'), so that we practically get just one sample. It should be always possible to implement a `num` argument in the `result` generator. Even if only `num = 1` would normally be desired. This could be supported by implementing default values for `num`, see e.g. `TimeConvolve`.

**Extra dimensions** in the output of `result`: there is already the `Filterbank` class, where this problem is solved by flattening all but the first (number of samples) dimensions so the output is always 2D. This approach ensures that the following block does not need to (but can) take care of extra dimensions.

**Incompatible input or outputs**: here I think a lean inheritance based approach, where we introduce some additional do-nothing base classes that allow to restrict which classes can serve as `source` for a certain class. This would need a careful inheritance design, but no actual implementation other than class declaration. Just to sketch the idea:

- Base class `Generator` (this would be the renamed `SamplesGenerator`)
- `SamplesGenerator` -> `Generator`
- `SpectraGenerator` -> `Generator` (maybe introduce an extra trait `numfreq`)
- `InOut` -> `Generator` (this woud be the renamed `TimeInOut`)
- `InTimeOut` -> `SamplesGenerator`/`InOut`
- `InSpectraOut` -> `SpectraGenerator`/`InOut`
- etc.
Actual processing classes would be derived from either `InTimeOut`, `InSpectraOut` or the like. The `source` trait will be restricted to `SamplesGenerator` and/or `SpectraGenerator` or some even more specialized type.

**Confusing class name where the user does not know what happens**: introduce alias names for some classes, e.g. `Average = TimeAverage`

esarradj:
> I added the new operators < and > to the TimeInOut and FreqInOut classes so that computational pipelines can be built not only by ..

While this looks kinda cool, the interpretation is not really intuitive to me. If we have
```
ta = ac.TimeSamples() > ac.TimePower() > ac.TimeAverage()
ta2 = ac.TimeAverage() < ac.TimePower() < ac.TimeSamples()
```
what is `ta` or `ta2`. An object of type `ac.TimeAverage` or `ac.TimeSamples` ? Or a generator for results? And how can I access say the `name` trait to set the input file? In most cases, one would have to assign values to traits of each of the processing blocks. This requires to use the traditional syntax anyway, so there is little to be gained here imho.

Regarding the suggested implementation, the operator based approach circumvents (?) the trait type mechanism and seems to need redundant type checks. If this is really necessary, then I think it is bad and a possible source of error if one make changes to the code.

PR #302, opened 2024-09-11
Refactor beamformer time classes

### LINKED ISSUE #299: DOCS: BeamformerTime.result() yields sound pressure time signal, NOT squared or deconvolved maps
### Type of Problem

Wrong Information: The documentation contains incorrect information.

### Documentation Issue

Doc of the result function says "Python generator that yields the squared deconvolved beamformer output with optional removal of autocorrelation block-wise." 

This is not correct. The yielded values are pure sound pressures, no deconvolution, no autocorrelation removal.

### Proposal for Change

Docstring "Python generator that yields the beamformer output (sound pressure time signals) block-wise." or something to that effect

adku1173:
Refactoring of `tbeamform` module:

* fixes wrong and missing doc strings (closes #299 )
* adds new `SamplesBuffer` class located in `tools.utils` for internal use (moves buffer logic away from Beamformer classes)
* extends BeamformerTime tests (additionally test result for last (shorter) output block)
* make only internally used methods private


FrankK20 (dismissed):
The `buffer` branch was tested with a test script and compared with the `master` branch. Various moving source configurations were analysed with the `BeamformerCleantTraj` (5 configurations in total, each calculated 10 times with `master` and `buffer` branch). The test looked at how well the source signal is reconstructed and the runtime. The test showed a comparable runtime between the `buffer` and the `master` branch and an identical reconstruction of the source signal. The screenshot shows the values for the runtime (average time) and the deviation in the signal reconstruction (average MSE) for the 5 cases for both branches (_buffer, _master). 
Screenshot from 2024-09-13 12-44-30

Additional tests were carried out with the `Blockwise_min.py` script in which several signal blocks of a moving source were reconstructed. Here too, a comparable runtime was observed between `master` and `buffer` branch and identical reconstructed signals. It was observed that the `acoular.tools.return_result` does not work as described in the `master` branch, but does in the new `buffer` branch. In the `master` branch, the call `acoular.tools.return_result(bcttc, num=1)`only returned one sample long signals, in the `buffer` branch all samples were returned. 

The scripts used for this tests  can be found in the attached scripts.zip file.

[scripts.zip](https://github.com/user-attachments/files/17013540/scripts.zip)

gherold (approved):

Issue #303, opened 2024-09-11
Remove *Sq classes


adku1173:
@esarradj, @gherold 

Is there anything against deprecating the `BeamformerTimeSq` class and all other `*Sq` classes in `acoular.tbeamform`? 
The `result` functions have already been merged anyway to a single function in `BeamformerTime` and `BeamformerTimeTraj` and we currently have 4 extra classes implementing the sq logic.
Instead, we could introduce a `squared_output` / `squared_result` or simply `sq` attribute, which is `False` by default. `r_diag` would then only be able to be used when `sq=True`. 

The only thing I am afraid of is that people think they apply "diagonal removal" but don't set the squared attribute to True. We could avoid this by changing the default value of `r_diag` to `False` and by printing a warning if `r_diag==True` and `sq==False`.

esarradj:
I think this change is unnecessary. The implementation of the *Sq classes is relatively lightweight and removing them has not much effect except breaking backwards compatibility.

artpelling:
Not intended.

PR #318, opened 2024-09-30
Enable to set steering vector type for all time domain beamforming classes

### LINKED ISSUE #314: BeamformerTime classes only use steering vector type 3
The four classes 
* `BeamformerTime`
* `BeamformerTimeSq`
* `BeamformerCleant`
* `BeamformerCleantSq`

only make use of steer type III (even if type I, II or IV is set). This should be settable similar to `BeamformerTimeTraj`.

adku1173:
* closes #314 
* adds private `_steer_funcs_freq` and `_steer_funcs_time` attributes to `SteeringVector`. This may be used in the future as a non-private attribute to inject custom steering vector functions.
* tests steering vector implementations


esarradj (dismissed):
There is no test yet for steering vector type. This should either be included here, or in any PR that addresses #303.

adku1173:
> There is no test yet for steering vector type. This should either be included here, or in any PR that addresses #303.

Tests are extended to all four steering vector formulations

esarradj (approved):

Issue #322, opened 2024-10-07
BUG: Numpy/OpenBlas UserWarning even when Acoular is loaded first


ti-sch:
### Version Dependencies

Python 3.11.9 and 3.12.7
Acoular 24.07
OS Debian

### Bug Description

When importing Acoular, the Acoular UserWarning 
`We detected that Numpy is already loaded and uses OpenBLAS. Because this conflicts with Numba parallel execution, we disable parallel execution for now and processing might be slower. To speed up, either import Numpy after Acoular or set environment variable OPENBLAS_NUM_THREADS=1 before start of the program.`
is shown.

This is done to prevent Thread Overloading (see #187)

In previous versions, this could be resolved by following the instructions and calling `import acoular` before importing numpy or any other package that uses it.
In version 24.07 the Warning is triggered either way, which also deactivates numbas parallelization by calling `numba.set_num_threads(1)` .

This has likely to do with the import order within Acoular itself, which seems to have changed: https://github.com/acoular/acoular/blob/b3e20f047e6f0b9197946099f45a6ae512781285/acoular/__init__.py#L9-L11

Both, `tools` and `calib` load numpy. 

Note that setting the environment variable does not avoid this either, since its value does not get checked by Acoular. 

### Possible Solution
- Changing the import order within [acoular/__init__.py](https://github.com/acoular/acoular/blob/e7f69857f4a05dbf7f71cfbe2572fa6cf9db829b/acoular/__init__.py) back to load `.configuration` first should resolve the issue.
- Checking `os.environ['OPENBLAS_NUM_THREADS']` before raising the warning, in case it was set by the user.

### Workaround
The code will work even with the warning, but performance might be lost. If you are sure that you are not overloading your threads, you can call `numba.set_num_threads(n)` after importing Acoular.

### Reproducible Code

```python
#simply:
import acoular
```

### Issue Severity

Medium: It affects my work, but I can find a workaround.

adku1173:
Thank you for reporting this bug. This behaviour was indeed not intended, and we will try our best to solve this issue in the upcoming release this month. Since you already have a plausible solution for the problem, I encourage you to become an Acoular contributor by submitting a pull request that includes the proposed changes. For details on how to contribute, see the [contribution guidelines](https://www.acoular.org/contributing/index.html)

ti-sch:
Thanks for the quick reply. If you agree that changing the import order is all it takes, that's simple enough.
I'm happy to create a Pull Request in that case.

Would you include a check?
[...]

PR #324, opened 2024-10-08
Fix UserWarning on import

ti-sch:
* Change import order in `acoular/__init__.py`

* Add check for OPENBLAS_NUM_THREADS


ti-sch:
PR fixes #322

ti-sch:
As described in the contribution checklist, I added myself to the CITATION.cff. This feels a bit silly for such a small change. Feel free to revert it.

adku1173:
> As described in the contribution checklist, I added myself to the CITATION.cff. This feels a bit silly for such a small change. Feel free to revert it.

Helping improve the package is nothing silly, and we are always happy to have new contributors. In your case, the change might be small but significant. We are hopeful that this change will not be your last.

adku1173 (approved):

PR #329, opened 2024-10-15
test docs on MacOs

artpelling:
This PR improves the CI for the docs:
- docs are now run for all platforms
- artifacts for deployment are only uploaded by `ubuntu-latest` as before
- tests now run on `macos-latest` rather than outdated `macos-13` (was pinned due to `pytables` headaches)
- we no longer install `graphviz` with the system package manager
- a dedicated CI action is now called for installing the `portaudio` system dependency which switches between `apt` and `brew` depending on the platform (windows works natively).

### DIFF
--- a/.github/actions/setup-conda/action.yml
+++ b/.github/actions/setup-conda/action.yml
@@ -4,7 +4,7 @@ inputs:
   python-version:
     description: 'The Python version'
     required: false
-    default: '3.12'
+    default: '3.13'
 runs:
   using: composite
   steps:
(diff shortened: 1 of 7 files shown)


artpelling:
rebased on `master` should be ready to go

esarradj (approved):

adku1173 (changes requested):
Can you add the PR description to the What's new internal section in the docs?

adku1173 (approved):

Issue #167, opened 2024-04-09
Numpy repmat is deprecated


esarradj:
Numpy `repmat` is deprecated from v1.19 and must be replaced by `tile`

artpelling:
closed by #330

PR #330, opened 2024-10-17
replace np.matlib.repmat with np.tile

gherold:
Solves #167


adku1173 (changes requested):
Thanks @gherold! Can you please add a short bullet point to the what's new section in the docs (see contribution checklist: https://www.acoular.org/contributing/checklist.html)?

adku1173 (approved):

PR #331, opened 2024-10-18
Numpy 2.0

adku1173:
Allow Numpy version > 2.0 for Python 3.8, 3.10, 3.11 and 3.12

CI only fails for Python 3.9 when Numpy > 2.0 due to missing the most recent Pytables version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 ]
 dependencies = [
     "numba",
-    "numpy<2.0",
+    "numpy",
     "scipy>=1.1.0",
     "scikit-learn",
     "tables>=3.4.4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,13 +28,11 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
-    "numpy<2.0; python_version == '3.9'",
-    "numpy; python_version == '3.8'",
-    "numpy; python_version >= '3.10'",
+    "numpy",
     "numba",
     "scipy>=1.1.0",
     "scikit-learn",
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -30,8 +30,10 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
+    "numpy<2.0; python_version == '3.9'",
+    "numpy; python_version == '3.8'",
```
artpelling, line 34:
```suggestion
    "numpy<2.0; python_version < '3.10'",
```

artpelling: maybe this is easier to read/fix later?

artpelling (approved):
Looks good. Maybe add a note to What's new?

pyproject.toml
adku1173: I suggest to first drop Python 3.8 and 3.9 and then fix the version requirements for tables and numpy

artpelling: Done, you can merge it :)

esarradj:
sources.py:83: DeprecationWarning: The numpy.linalg.linalg has been made private and renamed to numpy.linalg._linalg. All public functions exported by it are available from numpy.linalg. Please use numpy.linalg.norm instead.

artpelling:
> sources.py:83: DeprecationWarning: The numpy.linalg.linalg has been made private and renamed to numpy.linalg._linalg. All public functions exported by it are available from numpy.linalg. Please use numpy.linalg.norm instead.

Since we use a mix of `numpy.linalg` and `scipy.linalg`, we might just want to switch to `scipy.linalg` anyway as it is mostly a superset of the former.

PR #332, opened 2024-10-18
drop support for Python 3.8 and 3.9

artpelling:
Seeing that dependency resolution for older `pytables` versions is giving us headaches and 3.8 is now officially EOL, I think it is good to drop support in the next release. Also dropping 3.9 would resolve all `pytables` related issues (for now). Moreover, all NumFocus projects that we rely on have dropped 3.9 support in the latest versions.
But I have a feeling that @esarradj would like too keep it :wink: 

~I've added support for 3.13 because I don't see any reason not to.~
I see [the reason](https://github.com/numba/numba/pull/9682) now ...

### DIFF
--- a/.github/actions/setup-conda/action.yml
+++ b/.github/actions/setup-conda/action.yml
@@ -4,7 +4,7 @@ inputs:
   python-version:
     description: 'The Python version'
     required: false
-    default: '3.12'
+    default: '3.13'
 runs:
   using: composite
   steps:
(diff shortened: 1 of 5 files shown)


esarradj:
Don't keep 3.9 for me ...
If there is a good reason (and the dependencies dropping 3.9 support is) then we can surely drop it. Who requires 3.9 can still use older Acoular versions.

adku1173 (approved):

Issue #214, opened 2024-04-09
Extensive code line length


adku1173:
Some files include very long lines of code (>120 characters), including:
* [fastFuncs.py](acoular/fastFuncs.py)
* [fbeamform.py](acoular/fbeamform.py)
* [nidaqimport.py](acoular/nidaqimport.py)
* [signals.py (line 343)](acoular/signals.py)
* [sources.py (line 521)](acoular/sources.py)
* [spectra.py (line 741)](acoular/spectra.py)
* ... etc.

What code line length shall we consider to be valid?

artpelling:
I think 120 for code and 100 for docstrings is nice. IMO, anything over 120 can definitely be considered too long but sometimes it is nice to be able to have a long line. For the docstrings, 100 is more on the long side but should keep them readable even in a split screen without scrolling.

PR #336, opened 2024-10-19
Reduce line length to 120

jtodev:
### Description
* reduces (almost) all lines to 120 characters or less

### Comments
I sometimes used the join method to split and recombine stings since simply adding two strings together with a "+" triggers another linting error.

I sometimes used "neqa: E501" since the docstring would have been broken upon splitting parameter datatype arguments.

### Issue
This PR is regarding issue #214.


artpelling:
Closing in favour of #354.

PR #351, opened 2024-11-11
Calib

### LINKED ISSUE #344: Derive Calib class from InOut or TimeOut
Currently, calibration values can only be inserted into the signal processing chain via the Calib class using the TimeSamples / `MaskedTimeSamples` classes. 
Unfortunately, using the `SoundDeviceSamplesGenerator` class to generate calibrated signals in a streaming application is impossible. This is problematic because devices such as the UMA16 do not deliver sound pressures but voltage values. There are many more situations where the calibration of a signal chain would be beneficial (e.g. using different sensor types).

I can think of two approaches to solve this problem:

1. add a sensitivity attribute in the `SoundDeviceSamplesGenerator` class

2. derive `Calib` class from `TimeOut` (or even `InOut`) so that it can be used as an additional processing block to handle calibration in time or frequency domain. 

I prefer option two and suggest deprecating the `calib` attribute in `TimeSamples` and `MaskedTimeSamples`.

FrankK20:
### Description

Changes the Calib class so that it can be used as a separate processing block to handle calibration. As a source a `SpectraGenerator` or `SamplesGenerator` object can be used. 
In detail:
- derives Calib class from InOut 
- adds comments 
- adds deprecation warnings in TimeSample class 

### Reference issue 
 
Closes #344 

### Checklist


acoular/calib.py
```diff
@@ -12,16 +12,24 @@
...
+
+from .base import TimeOut
 
 # acoular imports
 from .internal import digest
 
 
-class Calib(HasPrivateTraits):
-    """Container for calibration data in `*.xml` format.
+class Calib(TimeOut):
```
adku1173, line 24: Calib should be better derived from `InOut` instead from `TimeInOut`

This enables the class to be used in the frequency domain. This should also be mentioned in the docstring (ideally with a short code snippet on how to use this class).

acoular/calib.py
```diff
@@ -12,16 +12,24 @@
...
+    """Container for calibration data in `*.xml` format or Numpy format (manually set).
+
+    This class loads calibration data and provides information about this data. It implemnts the
```
adku1173, line 27: implements

acoular/sources.py
```diff
@@ -326,6 +326,12 @@ def result(self, num=128):
         self._datachecksum  # trigger checksum calculation # noqa: B018
         i = 0
         if self.calib:
+            warn(
+                'The use of the calibration functionality in TimeSamples is deprecated and will be removed in \
+                       Acoular XX.XX. Use the Calib class as an additional processing block instead.',
```
adku1173, line 331: Deprecation date is 3 releases after the warning was introduced (25.10)

acoular/calib.py
```diff
@@ -76,3 +88,29 @@ def import_data(self):
...
+
+        Yields
+        ------
+        numpy.ndarray
+            Two-dimensional output data block of shape (num, numchannels)
+        """
+        for block in self.source.result(num):
+            if self.data.shape[0] == self.num_mics:
+                if self.num_mics == self.numchannels:
+                    yield block * self.data[newaxis]
```
adku1173, line 111: This line will only work in the time domain. If the source is of type `SpectraGenerator`, the data will is of shape `(num, numchannels*numfreq)` or even `(num, numchannels*numchannels*numfreq)`. A unittest that ensures the correct behaviour would be beneficial for this class.

adku1173 (changes requested):

acoular/calib.py
```diff
@@ -12,18 +12,62 @@
...
+        >>> print(next(fft.result(num=1)))
+        [[ 0.21277598+0.j          0.06519151-0.03153052j -0.04546857-0.06782166j
+           0.00149169+0.00993159j  0.03545392+0.07844662j  0.06390865+0.00359771j
+           0.00675808-0.07866847j  0.01878985+0.05065722j  0.05947051+0.j        ]]
+
+        One could also apply the calibration after the FFT calculation.
+
+        >>> fft = ac.RFFT(source=ts, block_size=16)
+        >>> calib = ac.Calib(source=fft)
+        >>> calib.data = 1e-2 * np.ones(ts.numchannels * fft.numfreqs)
```
adku1173, line 64: 100 instead of 1e-2

acoular/calib.py
```diff
@@ -32,15 +76,32 @@ class Calib(HasPrivateTraits, DeprecatedFromFile):
     file = File(filter=['*.xml'], exists=True, desc='name of the xml file to import')
 
     #: Number of microphones in the calibration data,
-    #: is set automatically / read from file.
+    #: is set automatically when read from file or when data is set.
     num_mics = CLong(0, desc='number of microphones in the geometry')
 
     #: Array of calibration factors,
-    #: is set automatically / read from file.
+    #: is set automatically when read from file can be set manually.
```
adku1173, line 83:
```suggestion
    #: is set automatically when read from file. 
    #: Can be set manually by specifying a NumPy array with shape (num_channels, ) if :attr:`source` yields
    #: time domain signals. For frequency domain signals, the expected shape is (num_channels * num_freqs).
```

acoular/calib.py
```diff
@@ -57,3 +118,29 @@ def import_data(self):
...
+        Yields
+        ------
+        numpy.ndarray
+            Two-dimensional output data block of shape (num, sourcechannels)
+            where sourcechannels is numchannels if the source data is in the time domain
+            or sourcechannels is numchannels*numfreqs of numchannels*numchannels*numfreqs
+            if the source data is in the frequency domain.
+        """
+        for block in self.source.result(num):
+            if self.data[self.channels].shape[0] == block.shape[1]:
```
adku1173, line 142: Can we check this before the generator starts? we could simply check if numchannels fits num_mics - len(invalid_channels)

acoular/calib.py
```diff
@@ -57,3 +118,29 @@ def import_data(self):
...
+                yield block * self.data[self.channels][newaxis]
+            else:
+                msg = f'calibration data shape does not match source data shape: {self.data.shape[0]}, {block.shape[1]}'
```
adku1173, line 145: The message does not account for invalid channels.

acoular/calib.py
FrankK20: The problem is that this would only work if the source is a time domain source. To account for frequency domain sources, we would need more logic.

adku1173 (approved):

Issue #204, opened 2024-04-09
Some solvers ind CMF can yield negative source strength results


adku1173:
including:

* FISTA
* Split-Bregman
* LassoLars
* LassoLarsBIC(?)

Reproduce with: 

```
import acoular
from acoular import L_p, Calib, MicGeom, Environment, PowerSpectra, \
RectGrid, MaskedTimeSamples,BeamformerCMF, SteeringVector

# other imports
from os import path
from pylab import figure, imshow, colorbar, title

# files
datafile = 'example_data.h5'
calibfile = 'example_calib.xml'
micgeofile = path.join( path.split(acoular.__file__)[0],'xml','array_56.xml')

#octave band of interest
cfreq = 4000

t1 = MaskedTimeSamples(name=datafile)
t1.start = 0 # first sample, default
t1.stop = 16000 # last valid sample = 15999
invalid = [1,7] # list of invalid channels (unwanted microphones etc.)
t1.invalid_channels = invalid 

t1.calib = Calib(from_file=calibfile)

m = MicGeom(from_file=micgeofile)
m.invalid_channels = invalid

g = RectGrid(x_min=-0.6, x_max=-0.0, y_min=-0.3, y_max=0.3, z=0.68,
             increment=0.05)
env = Environment(c = 346.04)
st = SteeringVector(grid=g, mics=m, env=env)
f = PowerSpectra(time_data=t1, 
               window='Hanning', overlap='50%', block_size=128, #FFT-parameters
               ind_low=8, ind_high=16) #to save computational effort, only
               # frequencies with indices 8..15 are used
bcmf = BeamformerCMF(freq_data=f, steer=st, method='LassoLars')

figure(1,(10,6))
smap = bcmf.synthetic(cfreq,1)
print(smap.min())
```

esarradj:
For sklearn solvers there is a new `positive` parameter which can be set to enforce strictly non-negative results.

PR #356, opened 2024-11-18
fix negative values

adku1173:
### Description

* Uses the `positive` argument in all `LassoLars` solvers to ensure positive results (sourcemaps)
* Adds a line in tests to check for negative values
* Unfortunately, `OrthogonalMatchingPursuitCV` leads to some negative values for the regression test when using `BeamformerDamasPlus` (`OrthogonalMatchingPursuitCV` has no attribute to ensure positivity)

### Reference issue

related to #204 

### Checklist


artpelling (dismissed):

artpelling:
@adku1173  tests are currently failing.

adku1173:
@artpelling I am confused. The problem is that the locally resulting source maps are slightly different from the CI source maps, although the algorithm's parametrization and the produced results are plausible. I already tried to pin the scikit-learn version without success. 

We could save the snapshots produced by the CI or expect the tests to fail without being strict, meaning that the snapshot test is allowed both to fail or pass for GIB with LassoLarsCV / LassoLars.

ci
local
diff

artpelling:
> @artpelling I am confused. The problem is that the locally resulting source maps are slightly different from the CI source maps, although the algorithm's parametrization and the produced results are plausible. I already tried to pin the scikit-learn version without success.
> 

Hmm thats really unfortunate. This can only be due to something happening under the hood.

> We could save the snapshots produced by the CI or expect the tests to fail without being strict, meaning that the snapshot test is allowed both to fail or pass for GIB with LassoLarsCV / LassoLars.
> 

Not sure what is best here. Could we maybe raise the tolerance without defeating the purpose of the test? Otherwise, I'd say we should expect it to fail or pass since that is the truth 😅 

@acoular/all Any other ideas?

esarradj:
I think the tests must give the same result (pass/fail) regardless of whether they are executed in the CI or not. Unless we do not find a definitive reason and actual failure, I would opt for relaxing the tolerance. See also this recommendation: https://github.com/scikit-learn/scikit-learn/discussions/27120

One thing that might be the actual reason for this behavior is that CV solvers pick some training data by splitting the available data. Maybe, the splitting process depends somehow on the hardware or the state of the random_number generator. The latter should not be the case, see https://scikit-learn.org/1.5/modules/generated/sklearn.model_selection.KFold.html#sklearn.model_selection.KFold , but who knows.

artpelling (approved):

PR #374, opened 2024-11-28
Test traits dependencies

### LINKED ISSUE #359: Anti pattern: properties that depend on independent properties
We should write a test to avoid the construct of properties that rely on other properties that do not depend on any other trait (mainly when calculating the digest!). The following example does not trigger any notification (same digest, although property value changes):

```Python
from acoular.internal import digest
from traits.api import Property, HasPrivateTraits, Int, cached_property 


class Test( HasPrivateTraits ):

    some_property = Property()

    _some_property = Int

    digest = Property( depends_on = ['some_property'] )

    def _get_some_property( self ):
        return self._some_property

    def _set_some_property( self, value ):
        self._some_property = value

    @cached_property
    def _get_digest(self):
        return digest(self)

if __name__ == "__main__":

    t = Test()
    t.some_property = 1
    print(t.digest)
    t.some_property = 2
    print(t.digest)

```

results in 

`_9d4ea2a37415bda19695e6210cf513a0
_9d4ea2a37415bda19695e6210cf513a0
`

Fortunately, such constructs hardly exist, but we should ensure they don't exist. Found it in 

* `LineGrid` (`size` trait)
* `MergeGrid` (`grid_digest` trait)
* `GenericSignalGenerator` (`numsamples` trait)
* `TimeSamples` (`_datachecksum` trait)
* `PowerSpectra` (`calib` trait)
* `SpatialInterpolator` (`mics_virtual` trait)

which indeed creates problems:

```Python
import acoular as ac
import numpy as np

mics_virtual = ac.MicGeom( mpos_tot = np.random.randn(3,3) )
print('mic digest', mics_virtual.digest)
interp = ac.SpatialInterpolator()
print('SpatialInterpolator digest', interp.digest)

# change the mics_virtual
mics_virtual.mpos_tot = np.random.randn(3,3)
print('mic digest', mics_virtual.digest)
print('SpatialInterpolator digest', interp.digest)
```

Output:
```
mic digest _996f9dfb5a9ea06ee04dab3144162104
SpatialInterpolator digest _383696c4c684315958bc46508cd29e8a
mic digest _40bd94e59ee802db8c5a22ca92d56fba
SpatialInterpolator digest _383696c4c684315958bc46508cd29e8a
```

artpelling:
### Description

- Adds a test that recursively ensures that any `Property` trait that is referenced by `depends_on` in another `Property` trait `depends_on` something.
- ~Refactors the `depends_on` keys to always be a list of strings. Before, `'a, b, c'` was used alongside `['a', 'b', 'c']` but now the latter is needed for the test to pass. Both are equivalent as per the [traits docs](https://docs.enthought.com/traits/traits_user_manual/listening.html#semantics).~
- Updates the `ruff` linting rules to deactivate `FIXME`
- Removes `noqa:TDXXX` and fixes them (proper `# TODO:` formatting)
- Turns `FlowField` into an abstract class, removes `FlowField.v` method, and removes `FlowField` from regression tests.
- New Issues #418, #419, #420 for remaining bugs the new test found

### Reference issue

Closes #359 

### Checklist

### DIFF
--- a/acoular/environments.py
+++ b/acoular/environments.py
@@ -16,6 +16,7 @@
 
 """
 
+from abc import abstractmethod
 import numba as nb
 from numpy import (
     arange,
@@ -47,7 +48,7 @@ from scipy.integrate import ode
 from scipy.interpolate import LinearNDInterpolator
 from scipy.linalg import norm
 from scipy.spatial import ConvexHull
-from traits.api import CArray, Dict, Float, HasStrictTraits, Instance, Int, Property, Union, cached_property
+from traits.api import ABCHasStrictTraits, CArray, Dict, Float, HasStrictTraits, Instance, Int, Property, Union, cached_property
 
 from .internal import digest
 
@@ -252,14 +253,16 @@ class UniformFlowEnvironment(Environment):
         return rm
 
 
-class FlowField(HasStrictTraits):
+class FlowField(ABCHasStrictTraits):
     """An abstract base class for a spatial flow field."""
 
     digest = Property
 
+    @abstractmethod
     def _get_digest(self):
-        return ''
+        pass
 
+    @abstractmethod
     def v(self, xx):  # noqa: ARG002
         """Provides the flow field as a function of the location. This is
         implemented here for the possibly most simple case: a quiescent fluid.
@@ -278,10 +281,6 @@ class FlowField(HasStrictTraits):
             - jacobian_matrix : array of floats of shape (3, 3)
                 The Jacobian matrix of the velocity vector field at the given location.
         """
-        v = array((0.0, 0.0, 0.0))
-        dv = array(((0.0, 0.0, 0.0), (0.0, 0.0, 0.0), (0.0, 0.0, 0.0)))
-        return -v, -dv
-
 
 class SlotJet(FlowField):
     """Provides an analytical approximation of the flow field of a slot jet.
--- a/acoular/environments.py
+++ b/acoular/environments.py
@@ -276,8 +276,7 @@ class FlowField(ABCHasStrictTraits):
 
     @abstractmethod
     def v(self, xx):  # noqa: ARG002
-        """Provides the flow field as a function of the location. This is
-        implemented here for the possibly most simple case: a quiescent fluid.
+        """Provides the flow field as a function of the location.
 
         Parameters
         ----------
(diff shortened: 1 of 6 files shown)


artpelling:
Currently waiting for a resolution of #375 before proceeding

artpelling:
@adku1173 I've removed `FlowField` from the regression test cases. Not sure if thats the way to go, though. If so, we might also have to delete some data?

acoular/environments.py
```diff
@@ -252,14 +265,16 @@ def _r(self, gpos, mpos=0.0):
...
 
+    @abstractmethod
     def _get_digest(self):
-        return ''
+        pass
 
+    @abstractmethod
     def v(self, xx):  # noqa: ARG002
         """Provides the flow field as a function of the location. This is
         implemented here for the possibly most simple case: a quiescent fluid.
```
adku1173, line 280:
```suggestion
        """Provides the flow field as a function of the location.
```

tests/unittests/test_classes.py
```diff
@@ -35,33 +38,86 @@ def test_instancing(acoular_cls):
...
+            tname = objtname[1]
+            check_or_unpack(obj, tname, toplevel=False)
+        else:
+            # deal with traits extended name definition by stripping trailing '[]'
+            # https://docs.enthought.com/traits/traits_user_manual/listening.html#semantics
+            tname = tname.replace('[]', '')
+            trait = obj.trait(tname)
+            if trait.is_property:
+                if trait.depends_on is None:
+                    assert toplevel, type(obj).__name__ + '.' + tname
```
adku1173, line 110: Better write a function that handles assert message formatting (function could be placed in /tests/utils.py).

adku1173 (changes requested):

acoular/environments.py
artpelling: @esarradj @gherold Should we add something like `ZeroFlowField` or similar for this functionality?

esarradj: This would only make sense for testing purposes, e.g. comparing the results of a. That's why I hid this in the abstract base class. A bit sloppy, I guess. If we do not want the abstract base class to implement something here, it could also be removed without any replacement imho.

adku1173 (approved):

esarradj (approved):

Issue #361, opened 2024-11-19
Inconsistent naming of attributes


adku1173:
Many variable names do not follow any concrete rules and are, therefore, chosen arbitrarily and inconsistently. This applies in particular to attributes that begin with num*:

* `num_mics`
* `num_sources`
* `num_blocks`
* `numfreqs`
* `numchannels`
* `numchannels_total`
* `numsamples`
* `numpoints`
* ...

Therefore, we should decide when to use underscores and when not to and extend the contribution section accordingly. Google (https://google.github.io/styleguide/pyguide.html#316-naming) and PEP8 recommend using underscores. This means that we should deprecate and rename variables accordingly (`num_samples` instead of `numsamples`, etc.).

artpelling:
`numpoints` is still undeprecated and `numsamples` is still present in `signals.py`. I haven't checked everything but we need to do a global search and replace on the entire repo.

PR #397, opened 2024-12-03
Update deprecation decorator

### LINKED ISSUE #388: Deprecated classes appear as LocalAliasHost objects
```Python
import acoular as ac
print(ac.RectGrid())
```

outputs:
```Python
<acoular.deprecation.deprecated_alias.<locals>.decorator.<locals>.LocalAliasHost at 0x78cc392fdcd0>
```

gherold:
### Description
Replaces some forgotten instances of deprecated traits "numsamples" and "numchannels" with their new counterparts (tests and examples are not yet updated!)

### Reference issue
Closes #388 and #361


artpelling (changes requested):
It would be good to move 71838a879fa644c642c2d8686f6554b097a83e51 to its own branch and start a separate PR to avoid issues with rebasing and make it easier to review. There a still plenty of instances of `numchannels`, `numpoints` etc. that need fixing, so that will result in a large diff.

adku1173:
Doctests fail. I suggest reverting the docstring formatting in the `Calib` class. I just recognized that the doctests never ran for the calibration module in #351 due to wrong formatting. @FrankK20 can you fix this?

adku1173:
We should first merge #398 and then this PR should remove the following two lines from the pyproject.toml file https://github.com/acoular/acoular/blob/f1bb2daa5cf27aa30f50d791c8966b243790fb24/pyproject.toml#L154-L155
which will cause the tests to fail if deprecated traits are used.

adku1173:
#398 has been merged. @gherold rebase is required.

adku1173:
@artpelling ready to review. I had a quick look into the build artifacts of the docs, which are correct.

acoular/deprecation.py
```diff
@@ -23,66 +23,63 @@ def deprecated_alias(old2new, read_only=False):
         If True, all deprecated traits will be "read only".
         If False (default), all deprecated traits can be read and set.
         If list, traits whose names are in list will be "read only".
+    removal_version: string or dict
+        Adds version of trait removal to deprecation message.
+        If not an empty string, it will be interpreted as version when
+        all traits in list will be deprecated.
+        If dictionary, keys are expected to be trait names and values
+        are the removal version strings.
     """
```
artpelling, line 32:
```suggestion
    removal_version: string or dict
        Adds the acoular version of trait removal to the deprecation message.
        If a non-empty string, it will be interpreted as the acoular version when
        all traits in the list will be deprecated.
        If a dictionary, the keys are expected to be trait names and the values
        are the removal version as strings.
    """
```

artpelling (dismissed):
Looks good so far :+1: 

However, there are is still use of deprecated traits in the `acoular/examples`. I found
- 11x `numsamples`
- 8x `numchannels`
- 3x `mpos`
- 4x `gpos_file`

and 3x `numsamples` in `acoular_demo.py`.

For consistency, I'd find it good to also rename the variables `numchannels` to `num_channels` in `tbeamform.py` and the same for `numpoints` in `fbeamform.py`. Although it is not strictly necessary.

artpelling:
I have replaced all remaining instances.

artpelling:
@gherold What's new section is still missing

adku1173 (approved):

PR #399, opened 2024-12-05
fix spelling mistakes

artpelling:


esarradj:
If we start correcting hyphens in compound nouns (I see no problem with that), we should maybe make clear in the contribution guidelines that we use AE rather than BE.

I got once corrected by an obviously British journal editor for my use of coordinate without hyphen. Admittedly this was some time ago.

artpelling:
> If we start correcting hyphens in compound nouns (I see no problem with that), we should maybe make clear in the contribution guidelines that we use AE rather than BE.
> 
> I got once corrected by an obviously British journal editor for my use of coordinate without hyphen. Admittedly this was some time ago.

I wasn't aware that that was BE spelling. I've reverted the changes. :gb:

adku1173 (approved):

Issue #216, opened 2024-04-10
Revise doc strings


adku1173:
Not all of the methods, modules, etc. contain docstrings (Linting fails for rules D100-D107), and some methods do not contain full argument descriptions (D417). In addition, all public classes should contain a short and extended summary line (D205).

- [x] `aiaa.py` -> #507 
- [x] `base.py` -> #507 
- [x] `calib.py` -> #507 
- [x] `deprecation.py` -> #507 
- [x] `environments.py` -> #444 
- [x] `fastFuncs.py` -> #507 
- [x] `fbeamform.py` -> #519 
- [x] `fprocess.py` -> #443 
- [x] `grids.py` -> #404 
- [x] `h5cache.py` -> #518 
- [x] `h5files.py` -> #517 
- [x] `internal.py` -> #507 
- [x] `microphones.py` -> #409 
- [x] `process.py` -> #442 
- [x] `sdinput.py` -> #507 
- [x] `signals.py` -> #431 
- [x] `sources.py` -> #415 
- [x] `spectra.py` -> #407 
- [x] `tbeamform.py` -> #515 
- [x] `tfastfuncs.py`
- [x] `tprocess.py` -> #457 
- [x] `traisviews.py` -> #507 
- [x] `trajectory.py` -> #414 
- [x] `version.py` -> #507

PR #404, opened 2024-12-20
docstrings for grids.py

jtodev:
### Description
New docstrings for grids.py

### Issue
#216

### Checklist

### DIFF
--- a/acoular/grids.py
+++ b/acoular/grids.py
@@ -2,7 +2,7 @@
 # Copyright (c) Acoular Development Team.
 # ------------------------------------------------------------------------------
 """
-Implements support for two- and threedimensional grids.
+Implement support for multidimensional grids and integration sectors.
 
 .. autosummary::
     :toctree: generated/
@@ -84,27 +84,31 @@ from .internal import digest, ldigest
 
 def in_hull(p, hull, border=True, tol=0):
     """
-    Test if points in `p` are in :attr:`hull`, in- or excluding the border.
+    Test if points in ``p`` are in ``hull``, in- or excluding the border.
 
     Parameters
     ----------
-    p : ndarray of floats, shape (N, K)
+    p : :class:`numpy.ndarray` of :class:`floats<float>`, shape `(N, K)`
         Coordinates of `N` points in `K` dimensions.
-    hull : ndarray of floats, shape (M, K), or :class:`scipy.spatial.Delaunay` object
+
+    hull : :class:`numpy.ndarray` of :class:`floats<float>`, shape `(M, K)`, or :class:`~scipy.spatial.Delaunay` object
         Coordinates of `M` points in `K` dimensions for which Delaunay triangulation will be
         computed.
+
     border : bool, optional
         Points in :attr:`p` on the border of :attr:`hull` will be kept in the return if `True`. If
         `False`, only points inside :attr:`hull` will be kept. Default is `True`.
-    tol : float, optional
-        Tolerance allowed in the inside-triangle check. Default is `0`.
+
+    tol : :class:`float`, optional
+        Tolerance allowed in the :meth:`inside-triangle check<scipy.spatial.Delaunay.find_simplex>`.
+        Default is ``0``.
 
     Returns
     -------
-    ndarray of bool
-        An array of boolean values indicating which points in :attr:`p` are inside the hull, same
-        shape as :attr:`p`. Each entry is `True` if the corresponding point is inside the hull (or
-        on the border, if `:attr:border=True`), and `False` otherwise.
+    :class:`numpy.ndarray` of :class:`bools<bool>`
+        An array of boolean values indicating which points in ``p`` are inside the hull, same
+        shape as ``p``. Each entry is ``True`` if the corresponding point is inside the hull (or
+        on the border, if ``border=True``), and ``False`` otherwise.
 
     Notes
     -----
@@ -121,7 +125,7 @@ def in_hull(p, hull, border=True, tol=0):
     >>> p = np.array([[0.5, 0.5], [2, 2]])
     >>> in_hull(p, hull)
     array([ True, False])
-    """
+    """  # noqa W505
     if not isinstance(hull, Delaunay):
         hull = Delaunay(hull)
 
@@ -140,11 +144,11 @@ def _det(xvert, yvert):
 
 class Polygon:
     """
-    Create a Polygon object representing a general polygon in a 2D plane.
+    Create an object representing a general polygon in a 2D plane.
 
-    This class allows defining a polygon by specifying the coordinates of its vertices
-    and provides methods for checking whether a point (or multiple points) lies inside
-    the polygon, or if a point is closer to a side or vertex of the polygon.
+    This class allows defining a polygon by specifying the coordinates of its vertices and provides
+    methods for checking whether a set of points lies inside the polygon, or if a point is closer to
+    a side or vertex of the polygon.
 
     Parameters
     ----------
@@ -158,11 +162,11 @@ class Polygon:
 
     Attributes
     ----------
-    x : ndarray
-        Array of x-coordinates of the polygon vertices, stored as a NumPy array.
+    x : :class:`numpy.ndarray`
+        Array of x-coordinates of the polygon vertices.
 
-    y : ndarray
-        Array of y-coordinates of the polygon vertices, stored as a NumPy array.
+    y : :class:`numpy.ndarray`
+        Array of y-coordinates of the polygon vertices.
     """
 
     def __init__(self, x, y):
@@ -184,28 +188,28 @@ class Polygon:
 
     def is_inside(self, xpoint, ypoint, smalld=1e-12):
         """
-        Check if a sigle point is or multiple points are inside a general polygon.
+        Check if a point or set of points are inside the polygon.
 
         Parameters
         ----------
-        xpoint : float or array_like
+        xpoint : :class:`float` or array_like
             Array of x-coordinates of the points to be tested.
 
-        ypoint : float or array_like
+        ypoint : :class:`float` or array_like
             Array of y-coordinates of the points to be tested.
 
-        smalld : float, optional
-            Small tolerance value used for floating point comparisons when checking if a point
-            is exactly on a polygon's side. The default value is `1e-12`.
+        smalld : :class:`float`, optional
+            Tolerance used for floating point comparisons when checking if a point is exactly on a
+            polygon's edge. The default value is ``1e-12``.
 
         Returns
         -------
-        float or array_like
+        :class:`float` or array_like
             The distance from the point to the nearest point on the polygon. The values returned
             have the following meanings:
-                - `mindst < 0`: Point is outside the polygon.
-                - `mindst = 0`: Point is on a side of the polygon.
-                - `mindst > 0`: Point is inside the polygon.
+                - ``mindst < 0``: Point is outside the polygon.
+                - ``mindst = 0``: Point is on an edge of the polygon.
+                - ``mindst > 0``: Point is inside the polygon.
 
         Notes
         -----
@@ -300,12 +304,11 @@ class Polygon:
 @deprecated_alias({'gpos': 'pos'})
 class Grid(ABCHasStrictTraits):
     """
-    Virtual base class for grid geometries.
+    Abstract base class for grid geometries.
 
-    This class defines a common interface for all grid geometries and provides tools to
-    query grid properties and related data. It is intended to serve as a base class for
-    specialized grid implementations and should not be used directly, as it lacks concrete
-    functionality.
+    This class defines a common interface for all grid geometries and provides tools to query grid
+    properties and related data. It is intended to serve as a base class for specialized grid
+    implementations and should not be instantiated directly as it lacks concrete functionality.
     """
 
     #: The total number of grid points. This property is automatically calculated based on other
@@ -316,8 +319,8 @@ class Grid(ABCHasStrictTraits):
     #: (read-only)
     shape = Property(desc='grid shape as tuple')
 
-    #: The grid positions represented as a (3, :attr:`size`) array of floats. Invalid points, such
-    #: as those corresponding to invalid microphones, are excluded. (read-only)
+    #: The grid positions represented as a (3, :attr:`size`) array of :class:`floats<float>`.
+    #: (read-only)
     pos = Property(desc='x, y, z positions of grid points')
 
     #: A unique identifier for the grid, based on its properties. (read-only)
@@ -353,14 +356,18 @@ class Grid(ABCHasStrictTraits):
 
         Parameters
         ----------
-        sector : :class:`Sector`
+        sector : :class:`Sector` object
             Sector describing the subdomain.
 
         Returns
         -------
-        tuple
-            A 2-tuple of arrays of integers or `numpy.slice` objects that can be used to
-            mask or select the specified subdomain from a grid-shaped array.
+        :class:`tuple`
+            A 2-tuple of arrays of integers or :obj:`numpy.s_` objects that can be used to mask or
+            select the specified subdomain from a grid-shaped array.
+
+        Notes
+        -----
+        The :func:`numpy.where` method is used to determine the the indices.
         """
         xpos = self.pos
         # construct grid-shaped array with "True" entries where sector is
@@ -378,22 +385,22 @@ class RectGrid(Grid):
     to the z-axis. It is defined by the lower and upper x- and y-limits and a constant z-coordinate.
     """
 
-    #: The lower x-limit that defines the grid. Default is -1.
+    #: The lower x-limit that defines the grid. Default is ``-1``.
     x_min = Float(-1.0, desc='minimum  x-value')
 
-    #: The upper x-limit that defines the grid. Default is 1.
+    #: The upper x-limit that defines the grid. Default is ``1``.
     x_max = Float(1.0, desc='maximum  x-value')
 
-    #: The lower y-limit that defines the grid. Default is -1.
+    #: The lower y-limit that defines the grid. Default is ``-1``.
     y_min = Float(-1.0, desc='minimum  y-value')
 
-    #: The upper y-limit that defines the grid. Default is 1.
+    #: The upper y-limit that defines the grid. Default is ``1``.
     y_max = Float(1.0, desc='maximum  y-value')
 
-    #: The constant z-coordinate of the grid plane. Default is 1.0.
+    #: The constant z-coordinate of the grid plane. Default is ``1.0``.
     z = Float(1.0, desc='position on z-axis')
 
-    #: The side length of each cell. Default is 0.1.
+    #: The side length of each cell. Default is ``0.1``.
     increment = Float(0.1, desc='step size')
 
     #: Number of grid points along x-axis. (read-only)
@@ -449,14 +456,14 @@ class RectGrid(Grid):
 
         Parameters
         ----------
-        x : float
+        x : :class:`float`
             The x coordinate of interest.
-        y : float
+        y : :class:`float`
             The y coordinate of interest.
 
         Returns
         -------
-        tuple of int
+        :class:`tuple` of :class:`ints<int>`
             Indices corresponding to the nearest grid point.
 
         Raises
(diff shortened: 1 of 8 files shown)


artpelling (changes requested):
Nice work so far, thanks! I've left some comments.

Any method starting with an underscore, i.e. `_get_foo` does not need a docstring. They are not rendered by sphinx. Especially all setters and getters of properties. It would be nice if you could remove them. This should also spare you some work going forward :D. If there is a private method that performs more involved computations, we could explain those briefly with a code comment.

artpelling (changes requested):

acoular/grids.py
```diff
@@ -647,23 +695,45 @@ def index(self, x, y, z):
         return xi, yi, zi
 
     def indices(self, x1, y1, z1, x2, y2, z2):
-        """Queries the indices for a subdomain in the grid.
+        """
+        Return the indices for a subdomain in the grid.
```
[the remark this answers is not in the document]
jtodev: I would rather leave it like this, since it's numpydoc standard to start the summary with an infinitive verb.

acoular/grids.py
```diff
@@ -256,55 +297,57 @@ def is_inside(self, xpoint, ypoint, smalld=1e-12):
...
     pos = Property(desc='x, y, z positions of grid points')
 
-    # internal identifier
+    #: A unique identifier for the grid, based on its properties. (read-only)
     digest = Property
 
     @abstractmethod
     def _get_digest(self):
-        """Returns the digest of the grid object."""
+        """Generate a unique digest for the grid."""
```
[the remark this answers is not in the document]
jtodev: Unfortunately it leads to `IndentationError`s if a method is defined without a proper definition below. A docstring prevents the `IndentationError` in such cases. So I suggest we keep them with the underscore methods in the `Grid` class.

acoular/grids.py
```diff
@@ -823,98 +967,129 @@ def _get_subgrids(self):
...
-        # subgrids = zeros((1,0))
         for grid in self.grids:
             bpos = append(bpos, grid.pos, axis=1)
-            # subgrids = append(subgrids,str(grid))
         return unique(bpos, axis=1)
 
 
 class Sector(ABCHasStrictTraits):
-    """Base class for all sector types.
+    """
```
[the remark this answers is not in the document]
jtodev: My mistake. I changed it.

acoular/grids.py
```diff
@@ -145,29 +181,34 @@ def __init__(self, x, y):
...
+        xpoint : float or array_like
+            Array of x-coordinates of the points to be tested.
 
-        xpoint and ypoint could be scalars or array-like sequences.
+        ypoint : float or array_like
+            Array of y-coordinates of the points to be tested.
 
-        Output parameters:
+        smalld : float, optional
+            Small tolerance value used for floating point comparisons when checking if a point
```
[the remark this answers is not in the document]
jtodev: I guess it's supposed to mean `Tolerance used`. I changed it to that.

acoular/grids.py
```diff
@@ -79,11 +81,44 @@
...
+    p : ndarray of floats, shape (N, K)
+        Coordinates of `N` points in `K` dimensions.
+    hull : ndarray of floats, shape (M, K), or :class:`scipy.spatial.Delaunay` object
+        Coordinates of `M` points in `K` dimensions for which Delaunay triangulation will be
+        computed.
+    border : bool, optional
+        Points in :attr:`p` on the border of :attr:`hull` will be kept in the return if `True`. If
+        `False`, only points inside :attr:`hull` will be kept. Default is `True`.
+    tol : float, optional
+        Tolerance allowed in the inside-triangle check. Default is `0`.
```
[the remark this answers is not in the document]
jtodev: All right. I got it to work and also applied it in a couple of other places.

acoular/grids.py
```diff
@@ -79,11 +81,45 @@
...
-    `p` should be a `NxK` coordinates of `N` points in `K` dimensions
-    `hull` is either a scipy.spatial.Delaunay object or the `MxK` array of the
-    coordinates of `M` points in `K`dimensions for which Delaunay triangulation
-    will be computed.
+    """
+    Test if points in ``p`` are in ``hull``, in- or excluding the border.
+
+    Parameters
+    ----------
+    p : ndarray of floats, shape `(N, K)`
```
adku1173, line 89: NumPy ndarrays should be referenced with 
```
:class:`numpy.ndarray` to be referenced by inter-sphinx.

File: acoular/grids.py [hunk 7]
```diff
@@ -79,11 +81,45 @@
...
+    border : bool, optional
+        Whether elements of ``p`` on the border of ``hull`` will be included. Default is
+        ``True``.
+    tol : float, optional
+        Tolerance allowed in the :meth:`inside-triangle check<scipy.spatial.Delaunay.find_simplex>`.
+        Default is ``0``.
+
+    Returns
+    -------
+    ndarray of bool
```
adku1173, line 103: Here, ndarray and bool should better be referenced:

```
:class:`numpy.ndarray` of :class:`bool`
```

acoular/grids.py
```diff
@@ -498,46 +535,61 @@ def indices(self, *r):
...
+        >>> grid.extend()
+        (-1.0, 1.0, -5.0, 5.0)
         """
         return (self.x_min, self.x_max, self.y_min, self.y_max)
 
 
 class RectGrid3D(RectGrid):
-    """Provides a cartesian 3D grid for the beamforming results.
+    """
+    Provides a cartesian 3D grid for the beamforming results.
```
artpelling, line 567: Then we need an infinite form here ;-)

acoular/grids.py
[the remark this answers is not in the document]
artpelling: I agree. I wasn't aware of this standard.

acoular/grids.py
```diff
@@ -672,29 +756,33 @@ def indices(self, x1, y1, z1, x2, y2, z2):
 
 @deprecated_alias({'from_file': 'file', 'gpos_file': 'pos'})
 class ImportGrid(Grid):
-    """Loads a 3D grid from xml file."""
+    """
+    Loads a 3D grid from an XML file.
```
artpelling, line 760: also here. Please check again if we use infinite form everywhere :)

artpelling (dismissed):

acoular/grids.py
```diff
@@ -256,67 +298,71 @@ def is_inside(self, xpoint, ypoint, smalld=1e-12):
...
 
-    #: Shape of grid. Readonly, gives the shape as tuple, useful for cartesian
-    #: grids
+    #: The shape of the grid, represented as a tuple. Primarily useful for Cartesian grids.
+    #: (read-only)
     shape = Property(desc='grid shape as tuple')
 
-    #: Grid positions as (3, :attr:`size`) array of floats, without invalid
-    #: microphones; readonly.
+    #: The grid positions represented as a (3, :attr:`size`) array of floats. Invalid points, such
```
adku1173, line 317: there is no functionality for invalid points. In addition, the current grid classes are not intended to be used as a microphone arrangement. Thus only `The grid positions represented as a (3, :attr:`size`) array of floats. (read-only)` makes sense

acoular/grids.py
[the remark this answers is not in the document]
adku1173: I don't see any problem with having docstrings for private methods. If a private method becomes public, it already has a proper docstring. In this case, the getter needs to be overwritten by corresponding subclasses, and some documentation can benefit new developers. @artpelling adding `pass` should not change functionality.

acoular/grids.py
```diff
@@ -498,46 +535,61 @@ def indices(self, *r):
...
     def extend(self):
-        """The extension of the grid in pylab.imshow compatible form.
+        """
+        Return the grid's extension in :obj:`matplotlib.pyplot.imshow` compatible form.
 
         Returns
         -------
-        4-tuple of floats
-            The extent of the grid as a tuple of x_min, x_max, y_min, y_max)
+        tuple of float
```
adku1173, line 543:
```suggestion
        tuple of floats
```

jtodev: How about linking to the classes in all cases? So in this case ``` :class:`tuple` of :class:`float` ``` or maybe ``` :class:`tuple` of :class:`floats<float>` ```. What do you think?

adku1173:
```
:class:`tuple` of :class:`floats<float>` 
```
looks good :)

acoular/grids.py
```diff
@@ -672,29 +756,33 @@ def indices(self, x1, y1, z1, x2, y2, z2):
 
 @deprecated_alias({'from_file': 'file', 'gpos_file': 'pos'})
 class ImportGrid(Grid):
-    """Loads a 3D grid from xml file."""
+    """
+    Loads a 3D grid from an XML file.
+
+    This class is used to import a 3D grid defined in an XML file. The grid's
```
adku1173, line 762: There is one crucial thing missing here. As with all classes supporting file import, there is a way of specifying a custom grid as a NumPy array, similar to the classes `MicGeom` and `TimeSamples`. For these classes, the documentation should contain an example demonstrating that setting the `pos` attribute is possible.

acoular/grids.py
```diff
@@ -611,22 +655,39 @@ def _get_digest(self):
...
+            array with the same shape as the grid.
+
+        Examples
+        --------
+        Check which of the points in a simple 8-point rectangular grid is closest to the point
+        ``(0.5, 0.5, 1.0)``.
+
+        >>> import acoular as ac
+        >>>
+        >>> grid = ac.grids.RectGrid3D()
```
adku1173, line 682: All classes in acoular can be directly imported (see the init.py file, which is not empty). Thus, it is always enough to write

```suggestion
        >>> grid = ac.RectGrid3D()
```

acoular/grids.py
```diff
@@ -647,23 +708,46 @@ def index(self, x, y, z):
...
+            same shape as the grid.
+
+        Examples
+        --------
+        Get the indices of the grid points of a simple 27-point rectangular grid which are located
+        inside the first octant.
+
+        >>> import acoular as ac
+        >>>
+        >>> grid = ac.grids.RectGrid3D()
```
adku1173, line 736:
```suggestion
        >>> grid = ac.RectGrid3D()
```

acoular/grids.py
```diff
@@ -708,7 +796,28 @@ def _set_pos(self, pos):
...
+        Notes
+        -----
+        The XML file should have elements with tag `pos`, where each `pos` element
+        contains attributes for `x`, `y`, `z`, and optionally `subgrid`.
+
+        Examples
+        --------
+        Consider two XML files, `grid1.xml` and `grid2.xml` containing different grids.
+
+        >>> from acoular.grids import ImportGrid  # doctest: +SKIP
```
adku1173, line 813: I prefer an example that writes a simple XML file (not with acoular since there is no `export_gpos` method) which is then loaded by the ImportGrid class. 

For consistency, I would prefer `import acoular as ac`

acoular/grids.py
```diff
@@ -498,46 +535,61 @@ def indices(self, *r):
...
+        Notes
+        -----
+        - ``pylab.imhow`` is the same as :obj:`matplotlib.pyplot.imshow`. It's only using a
+          different namespace.
+        - The return of the method is ment for the ``extent`` parameter of
+          :obj:`matplotlib.pyplot.imshow`.
+
+        Examples
+        --------
+        >>> grid = RectGrid()
```
adku1173, line 556: Why is no acoular import necessary here?

acoular/grids.py
```diff
@@ -721,28 +830,57 @@ def import_gpos(self):
...
+
+    #: Starting point of the grid in 3D space. Default is (0.0, 0.0, 0.0).
     loc = Tuple((0.0, 0.0, 0.0))
 
-    #: Vector to define the orientation of the line source
+    #: A vector defining the orientation of the line in 3D space. Default is (1.0, 0.0, 0.0).
     direction = Tuple((1.0, 0.0, 0.0), desc='Line orientation ')
 
-    #: Vector to define the length of the line source in meter
+    #: Total length of the line in meters. Default is 1.0.
```
adku1173, line 869: The values are not necessarily in meters. It depends on what the user defines as the unit of length.
```suggestion
    #: Total length of the line. Default is 1.0.
```

acoular/grids.py
```diff
@@ -721,28 +830,57 @@ def import_gpos(self):
...
-    #: Overall number of grid points. Readonly; is set automatically when
-    #: other grid defining properties are set
+    #: The total number of grid points. Automatically updated when other grid-defining attributes
+    #: are set. (read-only)
     size = Property(desc='overall number of grid points')
 
-    #: Grid positions as (3, :attr:`size`) array of floats, without invalid
-    #: microphones; readonly.
+    #: A (3, :attr:`size`) array containing the x, y, and z positions of the grid points, without
+    #: invalid microphones. (read-only)
```
adku1173, line 880: It's a source grid, not a microphone geometry (no invalid microphones).

acoular/grids.py
```diff
@@ -721,28 +830,57 @@ def import_gpos(self):
...
+    - The distance between points is :attr:`length` ``/ (`` :attr:`num_points` ``- 1)``.
+    - The direction vector is normalized to ensure consistency.
+
+    Examples
+    --------
+    Create a line grid with 5 points along the x-axis, starting at (0, 0, 0), with a length of 4
+    meters:
+
+    >>> import acoular as ac
+    >>> grid = ac.grids.LineGrid()
```
adku1173, line 852:
```suggestion
    >>> grid = ac.LineGrid()
```

acoular/grids.py
```diff
@@ -775,21 +910,47 @@ def _get_pos(self):
...
+    Notes
+    -----
+    - The merged grid eliminates duplicate points based on their positions.
+    - Each subgrid retains its original grid properties, such as digest and size.
+
+    Examples
+    --------
+    Merging two simple grids:
+
+    >>> from acoular import grids
```
adku1173, line 929: please use `import acoular as ac`

acoular/grids.py
```diff
@@ -823,98 +981,129 @@ def _get_subgrids(self):
...
-    #: Absolute tolerance for sector border
+    #: The absolute tolerance to apply when determining if a grid point lies on the sector border.
+    #: Default is `1e-12`. (optional)
     abs_tol = Float(1e-12, desc='absolute tolerance for sector border')
 
-    #: Boolean flag, if 'True' (default), the nearest grid point is returned if None is inside the
-    #: sector.
+    #: If `True`, the nearest grid point is returned if no grid points are inside the sector.
+    #: Default is `True`. (optional)
     default_nearest = Bool(True, desc='return nearest grid point to center of none inside sector')
```
adku1173, line 1058: There is a typo, but even when fixing this, the sphinx docstring and trait description are unclear. `default_nearest` is an attribute, not a function. Thus, it can not return anything. Instead, it should mention its purpose and the method (`contains`) where the attribute becomes relevant. 

```suggestion
    default_nearest = Bool(True, desc='return nearest grid point to center if none inside sector')
```

acoular/grids.py
```diff
@@ -823,98 +981,129 @@ def _get_subgrids(self):
...
+    specified :attr:`x_min`, :attr:`x_max`, :attr:`y_min`, and :attr:`y_max` positions, defining the
+    lower and upper bounds of the rectangle along the x and y axes.
     """
 
-    #: The lower x position of the rectangle
+    #: The minimum x position of the rectangle. Default is `-1.0`.
     x_min = Float(-1.0, desc='minimum x position of the rectangle')
 
-    #: The upper x position of the rectangle
+    #: The minimum x position of the rectangle. Default is `-1.0`.
```
adku1173, line 1074:
```suggestion
    #: The maximum x position of the rectangle. Default is `1.0`.
```

acoular/grids.py
```diff
@@ -823,98 +981,129 @@ def _get_subgrids(self):
...
-            Array indicating which of the given positions lie within the
-            given sector
-
+        array of bools, shape (N,)
+            Array indicating which of the given positions lie within the given sector.
+
+        Examples
+        --------
+        >>> import acoular as ac
+        >>> grid = ac.grids.RectGrid(increment=2)
```
adku1173, line 1103:
```suggestion
        >>> grid = ac.RectGrid(increment=2)
```

acoular/grids.py
```diff
@@ -823,98 +981,129 @@ def _get_subgrids(self):
...
+        >>> sec = ac.grids.RectSector(x_min=0, y_min=0)
```
adku1173, line 1104:
```suggestion
        >>> sec = ac.RectSector(x_min=0, y_min=0)
```

acoular/grids.py
```diff
@@ -950,35 +1139,47 @@ def contains(self, pos):
...
-            given sector
-
+        array of bools, shape (N,)
+            A boolean array indicating which of the given positions lie within the cuboid sector.
+            `True` if the grid point is inside the cuboid, otherwise `False`.
+
+        Examples
+        --------
+        >>> import acoular as ac
+        >>> grid = ac.grids.RectGrid3D(increment=2)
```
adku1173, line 1179:
```suggestion
        >>> grid = ac.RectGrid3D(increment=2)
```

acoular/grids.py
```diff
@@ -950,35 +1139,47 @@ def contains(self, pos):
...
+        >>> sec = ac.grids.RectSector3D(x_min=0, y_min=0, z_min=0)
```
adku1173, line 1180:
```suggestion
        >>> sec = ac.RectSector3D(x_min=0, y_min=0, z_min=0)
```

acoular/grids.py
```diff
@@ -1020,39 +1221,56 @@ def contains(self, pos):
 
 
 class CircSector(SingleSector):
-    """Class for defining a circular sector.
+    """
+    Class for defining a circular sector.
 
-    Can be used for 2D Grids for defining a circular sector or
-    for 3D grids for a cylindrical sector parallel to the z-axis.
+    This class inherits from :class:`SingleSector` and defines a circular sector, which can be used
```
adku1173, line 1227:
```suggestion
    Defines a circular sector, which can be used
```

acoular/grids.py
```diff
@@ -1020,39 +1221,56 @@ def contains(self, pos):
...
-            given sector
-
+        array of bools, shape (N,)
+            A boolean array indicating which of the given positions lie within the circular sector.
+            `True` if the grid point is inside the circular sector, otherwise `False`.
+
+        Examples
+        --------
+        >>> import acoular as ac
+        >>> grid = ac.grids.RectGrid(increment=1)
```
adku1173, line 1266:
```suggestion
        >>> grid = ac.RectGrid(increment=1)
```

acoular/grids.py
```diff
@@ -1020,39 +1221,56 @@ def contains(self, pos):
...
+        >>> grid.pos
+        array([[-1., -1., -1.,  0.,  0.,  0.,  1.,  1.,  1.],
+               [-1.,  0.,  1., -1.,  0.,  1., -1.,  0.,  1.],
+               [ 1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.]])
+        >>> sec = ac.grids.CircSector(x=1, y=1, r=0.5)
```
adku1173, line 1271:
```suggestion
        >>> sec = ac.CircSector(x=1, y=1, r=0.5)
```

adku1173:
@jtodev  It would be nice to include a reference to the sector integration example for the sector classes. This should be possible using the `:ref:` directive for the link `.. _sphx_glr_auto_examples_wind_tunnel_examples_example_sectors_and_integration.py:` pointing to the `acoular/docs/source/auto_examples/wind_tunnel_examples/example_sectors_and_integration.rst` example

acoular/grids.py
jtodev: I don't know. `hatch run tests:doctest` works. But I will add `from acoular import RectGrid` out of rigorousness.

acoular/grids.py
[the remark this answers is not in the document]
jtodev: So I will just leave the docstrings there and make no further changes, right?

acoular/grids.py
```diff
@@ -823,98 +981,129 @@ def _get_subgrids(self):
...
-    directly, it implements a sector encompassing the whole grid.
+    Defines the common interface for all single sector classes. This class can serve as a base for
+    various single sector implementations. When used directly, it defines a sector that encompasses
+    the whole grid. It includes attributes for handling border inclusion, tolerance for sector
+    borders, and default behavior when no points are inside the sector.
     """
 
-    #: Boolean flag, if 'True' (default), grid points lying on the sector border are included.
+    #: If `True`, grid points lying on the sector border are included in the sector. Default is
+    #: `True`. (optional)
```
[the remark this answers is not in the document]
jtodev: Yes, I figured because of the default values. I removed it.

adku1173 (approved):

adku1173 (changes requested):
@jtodev Unfortunately, the link to the sector integration example does not work as expected (-> it is not shown after building the docs).

adku1173 (approved):

PR #411, opened 2025-01-15
Intersphinx

adku1173:
adds intersphinx connection for Python, SciPy and NumPy documentation.

One can now use for example 
```
:class:`numpy.ndarray`
``` 
for cross-linking, which then renders as:

image


artpelling (approved):
Maybe add a small section to the contributing section?

PR #416, opened 2025-01-21
Add new NoiseGenerator and PeriodicSignalGenerator base classes

### LINKED ISSUE #385: Bug: UncorrelatedNoiseSource needs revision
class `UncorrelatedNoiseSource` has some issues:

1. the dependencies in the `digest` property are missing the `seed` of the noise signal. The `digest` does not change when the `signal` changes its `seed`. Thus, the digest should better depend on `signal.digest`
2. `UncorrelatedNoiseSource` allows an Instance of `SignalGenerator`. One could also use a `SineGenerator` class. It would be better to have a `NoiseGenerator` base class which implements the `rms` attribute and from which `WNoiseGenerator` etc. are derived.

adku1173:
### Description
* Adds two new base classes to improve the inheritance hierarchy in signals.py 
* `UncorrelatedNoiseSource` now needs a signal derived from `NoiseGenerator` (instead of `SignalGenerator`, which falsely would allow passing SignalGenerator` objects)

### Reference issue
#closes #385


FrankK20 (approved):
Looks good.

acoular/signals.py
```diff
@@ -89,16 +88,70 @@ def usignal(self, factor):
         return resample(self.signal(), factor * self.num_samples)
 
 
-class WNoiseGenerator(SignalGenerator):
-    """White noise signal generator."""
+class PeriodicSignalGenerator(SignalGenerator):
+    """
+    Virtual base class for periodic signal generators.
```
artpelling, line 93:
```suggestion
    Abstract base class for periodic signal generators.
```

artpelling (changes requested):

adku1173:
@artpelling should be ready to merge now

artpelling (approved):

PR #422, opened 2025-01-23
Merge grid

### LINKED ISSUE #419: `MergeGrid` digest mechanism does not work as intended
### Version Dependencies

Python 3.12, acoular 24.10

### Bug Description

`MergeGrid.digest` depends on `MergeGrid.grid_digest` which is a `Property` that itself does not depend on anything. This will prevent the digest mechanism to be invoked, if any of the grids digests are changed.

### Reproducible Code

```python

```

### Error message:

```shell

```

### Issue Severity

None

adku1173:
closes #419

### DIFF
--- a/acoular/grids.py
+++ b/acoular/grids.py
@@ -65,17 +65,18 @@ from traits.api import (
     Int,
     List,
     Property,
+    Str,
     Tuple,
     Union,
     cached_property,
     on_trait_change,
     property_depends_on,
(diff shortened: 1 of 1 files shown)


artpelling:
@adku1173 please remove `MergeGrid` from the xfails list in `test_traits_dependencies`

artpelling (approved):

Issue #373, opened 2024-11-27
Consider switching to right-handed coordinate system


gherold:
For historical reasons, Acoular currently uses a left-handed coordinate system for handling spatial information (grid definitions, microphone array geometries, flow properties, ray tracing, ...).

Since most of the world (including other python packages) by convention uses right-handed coordinates, we should consider making these the default in Acoular as well. This would mean a major compatibility break with older scripts in a lot of cases, but the benefits for Acoular in terms of interoperability and intuitive usage also weigh heavy in my opinion.

We could either

1. completely switch to right-handed coordinates or
2. have a flag that can be set to RH or LH coords in the config

Benefit of 2. would be that this allows easy support for older scripts using Acoular. Benefit of 1. would be keeping a better maintainability of the code base.

esarradj:
In the `environment` module, this affects

- [ ] UniformFlowEnvironment (?)
- [ ] Slot Jet
- [ ] RotatingFlow
- [ ] GeneralFlowEnviroment (?)

In the `tbeamform` module this affects 

- [ ] BeamformerTimeTraj (?)

Interestingly I could not find any other code, but I might be wrong.

esarradj:
There is another situation which might be the basis of all the confusion about Acoular using a left-handed system, which actually not happens inside Acoular, but only when plotting or interpreting the results as e.g. shown [here](https://www.acoular.org/auto_examples/introductory_examples/example_3D_beamforming.html#sphx-glr-auto-examples-introductory-examples-example-3d-beamforming-py). The very same Acoular code could be used with a right handed co-ordinate system. This would mean to define the z-extension of the grid with negative z-values. The only thing that needs to be changed then are the plotting commands! But not Acoular, obviously. Do you agree, @gherold @adku1173 ?

gherold:
Yes, for most (if not all cases), using a left-oriented system now appears like a self-imposed obstacle not required by the original implementation. This is actually a most desirable outcome, since no compatibility-breaking changes are required and everybody can choose whichever system they like better (as long as they use it consistently).

esarradj:
I was under the impression that there would be some change to the cross product when using LHS or RHS. Even this is [not the case](https://stackoverflow.com/questions/4124041/is-opengl-coordinate-system-left-handed-or-right-handed?rq=3).

PR #427, opened 2025-01-24
Right hand co-ordinate system in examples

esarradj:
### Description
This PR updates most of the examples to use a right-hand co-ordinate system. From the Acoular legacy it was custom to use a left-handed system. However, the Acoular code itself is largely agnostic to the co-ordinate system used. Therefore, most of the examples can very easily be changed to right-handed, most often with the change of one or two minus signs.

### Reference issue
Contributes to the solution of #373.

### Checklist


artpelling:
Does it update _most_ examples in the sense that there are none left that use an LHS? If not, we should track the ones that are still to be done.

If dot products are unaffected, does this then close #373 ?

esarradj:
There was no change necessary to Acoular code, but only to the examples. This is ready for review now and could eventually be merged.

adku1173 (changes requested):
Very nice :). For consistency, I would prefer to always have `import matplotlib.pyplot as plt` (see: #432 ), and there is one issue with the figure numbers in the rotating source example (which will be solved with #433).

adku1173 (approved):

Issue #481, opened 2025-06-19
BUG: ``BeamformerBase.integrate`` returns empty array


jtodev:
### Version Dependencies

Python 3.12.11
Acoular 25.04
macOS 14.7.6 (23H626)

### Bug Description

**Context**
I am writing an example for sector integration in which I want to integrate over a sector covering the trailing edge of an airfoil in an wind tunnel using the ``BeamformerBase.integrate`` method with the said sector as input.

**Expected behavior**
I expect the ``BeamformerBase.integrate`` method to return the spectrum for the integrated sector.

**What actually happened**
The method return an empty array.

**Peculiarities**
Upon running ``bb.integrate(sector)`` a second time, the spectrum is returned.

### Reproducible Code

```python
import acoular as ac
from pathlib import Path

micgeofile = Path(ac.__file__).parent / 'xml' / 'array_64.xml'
mg = ac.MicGeom(file=micgeofile)
pa = ac.demo.create_three_sources(mg)
ps = ac.PowerSpectra(source=pa, block_size=128, window='Hanning')
rg = ac.RectGrid(x_min=-0.2, x_max=0.2, y_min=-0.2, y_max=0.2, z=-0.2, increment=0.01)
st = ac.SteeringVector(mics=mg, grid=rg)
bb = ac.BeamformerBase(freq_data=ps, steer=st)

sector = ac.RectSector(x_min=-0.09, x_max=-0.11, y_min=-0.09, y_max=-0.11)

bf_sector = bb.integrate(sector)
print(bf_sector)
```

### Error message:

```shell
[('Mixer_34777131016c0ccb75d8d0cc70ce56c1_cache.h5', 1)]
[('Mixer_34777131016c0ccb75d8d0cc70ce56c1_cache.h5', 2)]
[]
```

### Issue Severity

High: It significantly affects my work, and I cannot find a workaround.

PR #493, opened 2025-07-08
modify test to catch fixed case

artpelling:
I looked into why we didn't catch #481 in our tests in the first place. It turns out that we always calculate the results beforehand and that the fixture scope is too broad such that the cached results from previous test runs mask this behaviour.

The test duration is now increased slightly. Not sure what we want to do here, it is the classical dilemma of fast vs. feasible...


esarradj:
I would rather not change the scope here. The problem in #481 would be still captured at least for the very first Grid/Sector combination tested, or not?

Issue #478, opened 2025-06-17
`WriteWAV` needs rework


artpelling:
There are several issues with `WriteWAV` that warrant a dedicated rework.

We should think about including a `result` method.

Issue #473, opened 2025-06-10
`WriteWAV.save` calls the source twice


esarradj:
`WriteWAV.save` calls the source twice: first to find the maximum value to determine scaling, second to actually save values. If there is no `TimeCache` used before, then this results in the repeated computation of whatever signals get stored (twice the computation time).

Possible solutions could be: 
- force caching 
- internal storage of results to prevent re-computation
- (possibility to) set the scaling factor from a trait (this has also the advantage of consisted scaling between multiple recordings)
- use another WAV format (float32 instead of int16)

esarradj:
One option to overcome the auto max. value problem, I thought of would be to first write to a hdf5-File (possibly by reusing the code from `WriteH5` and after writing that file is complete to convert it to wav. 

I do not think that float32 .wav files are that important. Such files are often not supported by other software, anyway.

artpelling:
> One option to overcome the auto max. value problem, I thought of would be to first write to a hdf5-File (possibly by reusing the code from `WriteH5` and after writing that file is complete to convert it to wav.

I don't really like that idea, depending on the data and hardware, writing to disk twice can be very costly. I have implemented it such that the results are internally cached. Of course, that could also lead to memory problems. However, in such an extreme case the user can set `max_val` to circumvent that.

esarradj:
I meant use `TimeCache` in `WriteWAV`, `WriteH5` doesn't need that.
+1 for using `Cache` instead

The behavior could be as follows: only if `max_val` is unset and the `source` is not a `Cache`, instead of the `source` a `Cache` is created that uses `source` as its `source`

```
self.source = Cache(source = self.source)
```

Issue #382, opened 2024-11-29
Generator derived classes without result method


adku1173:
Some classes in Acoular violate the Generator principle by not implementing a dedicated result method. This includes 

* `Trigger`
* `SpatialInterpolator`
* `WriteWAV`

PR #498, opened 2025-07-17
rework `WriteWAV`

### LINKED ISSUE #126: WriteWAV: add custom normalization
currently, WriteWAV's save function automatically determines the scaling of the values. It should be possible to set the scaling externally, so that several wav files have the same normalization and can be compared. For this, an additional trait ("max_val" or similar) could be introduced, which normalizes all wav values (and clips if necessary)

artpelling:
### Description

- Implements `max_val` attribute that can be used to achieve consistent scaling.
- Avoids double computation of `source.result` by storing the result in the case that `max_val` is `None` and the maximum needs to be computed.
- Adds a warning, if `sample_freq` is not a whole number. Before, it was silently rounded in the `wave` module.
- Adds `encoding` attribute to allow also `uint8` and `int32` PCM formats. `uint24` would be straightforward to add in the future but I would not add it unless there is serious demand by the user base.
- now adds a `result` generator to `WriteWAV` that writes to file and passes through the data.
- removes `WriteWAV` from the blacklist in `test_result_generator`
- adds a warning if clipping occurs

### Reference issue

Closes #126 #478 #473 
Related #382 

### Checklist

### DIFF
--- a/acoular/tprocess.py
+++ b/acoular/tprocess.py
@@ -1822,12 +1822,29 @@ class WriteWAV(TimeOut):
 
     def _encode(self, data):
         """Encodes the data according to self.encoding."""
-        dtype, min_val, max_val, _ = self._type_info()
+        dtype, dmin, dmax, _ = self._type_info()
         if dtype == np.dtype('uint8'):
-            data = (data + 1) / 2 * max_val
+            data = (data + 1) / 2 * dmax
         else:
-            data *= -min_val
-        return np.round(data).clip(min_val, max_val).astype(dtype).tobytes()
+            data *= -dmin
+        data = np.round(data)
+        if data.min() < dmin or data.max() > dmax:
+            warn(
+                f'Clipping occured in WAV export. Data type {dtype} cannot represent all values in data. \
+            Consider raising max_val.',
+                stacklevel=1,
+            )
+        return data.clip(dmin, dmax).astype(dtype).tobytes()
+
+    def _compute_max_val(self, res):
+        """Computes the maximum value from the result data such that maximal scaling is achieved."""
+        dtype, _, dmax, _ = self._type_info()
+        if dtype == np.dtype('uint8'):
+            mx = abs(res).max()
+        elif dtype in (np.dtype('int16'), np.dtype('int32')):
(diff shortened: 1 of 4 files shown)


artpelling:
@acoular/acoular-development-team We need to resolve the discussion in #473 to decide how to proceed here.

acoular/tprocess.py
```diff
@@ -1807,38 +1815,90 @@ def _get_basename(self):
...
-        msg = 'result method not implemented yet! Data from source will be passed without transformation.'
-        warn(msg, Warning, stacklevel=2)
-        yield from self.source.result(num)
+                def result(num):
+                    yield from self.source.result(num)
+
+                mx = self.max_val
+            else:
+                # compute maximum and remember result to avoid calling source twice
+                res = return_result(self.source, num=num)
```
gherold: It is risky to put the data of the whole source into memory like this, since this would be num_samples x num_channels and could well be too large to handle for RAM (think time domain beamformer result). I see three options: 
1) If no max_val is given, keep block-wise calculation and have an option to put the data through a TimeCache object to avoid double calculation
2) use MaskedTimeInOut together with return_result to keep only the channels that you want to write to disk in memory.
3) Output a warning that no max_val is set and do the double calculation.

acoular/tprocess.py
```diff
@@ -1807,38 +1815,90 @@ def _get_basename(self):
...
+        """Encodes the data according to self.encoding."""
+        dtype, dmin, dmax, _ = self._type_info()
+        if dtype == np.dtype('uint8'):
+            data = (data + 1) / 2 * dmax
+        else:
+            data *= -dmin
+        data = np.round(data)
+        if data.min() < dmin or data.max() > dmax:
+            warn(
+                f'Clipping occured in WAV export. Data type {dtype} cannot represent all values in data. \
```
gherold: Consider adding the clipped values as info for the user so they can change max_val accordingly.

artpelling: that is hard to do currently because the warning will only be emitted once.

acoular/tprocess.py
artpelling: Let's discuss this together in #473 !

PR #512, opened 2025-07-28
fix #153

### LINKED ISSUE #153: Replace on_trait_change by observe method
From traits documentation:

Traits 6.1 introduces a new API for configuring traits notifications: observe, which is intended to fully replace an older API (on_trait_change) in order to overcome some of its limitations and flaws. While on_trait_change is still supported, it may be removed in the future. 

Since there is no deprecation warning this is nonurgent.

artpelling:
### Description
This pull request replaces the deprecated `on_trait_change` decorator with the `observe` decorator across multiple files in the codebase to modernize the code and ensure compatibility with the latest version of the Traits library. Below is a categorized summary of the changes:

### Reference issue

Closes #153 

### Checklist

### DIFF
--- a/acoular/aiaa/aiaa.py
+++ b/acoular/aiaa/aiaa.py
@@ -116,7 +116,7 @@ class CsmAIAABenchmark(PowerSpectraImport):
         return get_file_basename(self.file)
 
     @observe('basename')
-    def load_data(self, event):  # noqa ARG002
+    def _load_data(self, event):  # noqa ARG002
         """Open the .h5 file and set attributes."""
         if self.h5f is not None:
             with contextlib.suppress(OSError):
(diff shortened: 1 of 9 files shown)


esarradj (dismissed):

artpelling:
@esarradj I don't enjoy the fact that we change the call signature of public methods. I guess all of the concerned methods should have been private to begin with. I would suggest to make them private and deprecate their usage. Should I do this or am I missing something?

artpelling:
> Indeed, all these methods should be private as they have no public meaning. So, renaming seems to be a good idea. The only thing is that they never should start with `_set_` or `_get_` because this interferes with how traits handles property getters and setters. Maybe we could call it `_update_eva_list` and `_update_num_mics`

How do you feel about deprecations? I wonder if these methods are explicitly called by any user, it would probably be ok to just rename without deprecation?

jake-is-ESD-protected (approved):
This is code, very good🙏

PR #532, opened 2025-09-11
change `PointSource.loc` to `CArray`

artpelling:
There is no reason for this to be a `Tuple`. Everywhere it is used, it is cast as an array. Changing the trait allows to remove all the reshaping and casting. Some local benchmarking also revealed that this slightly speeds up things.

### DIFF
--- a/acoular/sources.py
+++ b/acoular/sources.py
@@ -720,8 +720,7 @@ class PointSource(SamplesGenerator):
     #: Instance of the :class:`~acoular.signals.SignalGenerator` class defining the emitted signal.
     signal = Instance(SignalGenerator)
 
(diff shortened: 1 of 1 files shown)


artpelling:
@esarradj Changing `PointSource.loc` to just a `CArray` does not accept any argument other than an array of the correct shape. You can maybe try for yourself by setting

``` python
loc = CArray(shape=(3, 1), value=np.array([[0.0], [0.0], [1.0]]), dtype=float)
```

I checked the [docs]*https://docs.enthought.com/traits/traits_api_reference/trait_numeric.html#traits.trait_numeric.CArray) and this should actually work, as you said. It was for this reason that I implemented it the way I did...

Issue #530, opened 2025-09-08
Add mode to `TimeConvolve.result`


jtodev:
In the current implementation of the `TimeConvolve` class's `result` method, the output size is specified as _L + M - 1_ with kernel size _L_ and _M_ samples. This does not necessarily correspond to the number of samples, i.e., _M_, which is the desired output size. To prevent this mismatch, the `result` method should be modified to include another mode to specify the output size.

The current implementation corresponds to the the `numpy.convolve` function's `'full'` mode. Given that _L < M_, a match between the sample and output size can be achieved by implementing the `'same'` mode, which sets the output size to _max(L, M)_. Another option for an alternative mode is the `'valid'` mode with output size _max(L, M) - min(L, M) + 1_.

For the `'same'` mode and `num=1`, I ran into the issue of the `result` method yielding empty last blocks.

Issue #514, opened 2025-07-31
BUG: `PointSourceConvolve.num_samples` does not match actual number of samples


jtodev:
### Version Dependencies

- Python 3.13.5
- Acoular 25.07
- macOS 14.7.7 (23H723)

### Bug Description

I'm currently revising the tests for the result generators in Acoular. One of the assertions I want to include is that the total number of samples produced by a generator's `result` method should match its `num_samples` attribute.

However, in the example below, this assertion fails: the signal is set to 50 samples, but the generator produces 54 samples in total.

After investigating, I found that this discrepancy originates from the `TimeConvolve` class's `result` method, where the output length is calculated as `input_length + kernel_length - 1`. For the case shown, this results in 54 samples.

Is this behavior intentional, or should the `num_samples` attribute reflect the actual output length after convolution?

### Reproducible Code

```python
import acoular as ac
from numpy import concatenate, array, zeros

micgeom = ac.MicGeom(pos_total=array([[0, 0, 0]]).T)
obj = ac.PointSourceConvolve(signal=ac.WNoiseGenerator(sample_freq=1000, num_samples=50, seed=1), mics=micgeom)

kernel = zeros(5)
kernel[1] = 1
obj.kernel = kernel

num = 32
blocks = list(obj.result(num))

result = concatenate(blocks, axis=0)
assert result.shape[0] == obj.num_samples, f'Total samples {result.shape[0]} != {obj.num_samples}'
```

### Error message:

```shell
Traceback (most recent call last):
  File "/Users/janostendorf/Desktop/test.py", line 15, in <module>
    assert result.shape[0] == obj.num_samples, f'Total samples {result.shape[0]} != {obj.num_samples}'
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Total samples 54 != 50
```

### Issue Severity

Low: It is a minor inconvenience but does not affect my work.

PR #534, opened 2025-10-01
Add modes to `TimeConvolve.result`

jtodev:
### Description
Based on the `numpy.convolve` function, the three modes `'full'`, `'same'`, and `'valid'` are added to the `TimeConvolve.result` method, corresponding to output sizes _L + M - 1_, _max(L, M)_, and _max(L, M) - min(L, M) + 1_, respectively, with kernel size _L_ and sample size _M_.

In the current implementation, the `'full'` mode is used. Prospectively, the implementation of one of the alternative modes will match the `result` method's output size to the sample size.

### Reference issue
#530 

### Checklist

- [ ] My changes fulfill the [Code Quality](https://www.acoular.org/contributing/quality.html) standards.
- [ ] I have updated the [What's new](https://www.acoular.org/news/index.html) section the the [documentation](https://github.com/acoular/acoular/blob/master/docs/source/news/index.rst) explaining my changes.


artpelling:
Looks good. Can you add tests for the modes and also ensure that #514 is fixed?

acoular/sources.py
```diff
@@ -2249,8 +2249,7 @@ class PointSourceConvolve(PointSource):
     (256, 4)
     (256, 4)
     (256, 4)
-    (256, 4)
-    (75, 4)
+    (232, 4)
```
artpelling, line 2252: why did you change this?

acoular/tprocess.py
```diff
@@ -2687,10 +2687,22 @@ class TimeConvolve(TimeOut):
 
     #: Convolution kernel in the time domain.
     #: The second dimension of the kernel array has to be either ``1`` or match
-    #: the :attr:`source`'s :attr:`~acoular.base.SamplesGenerator.num_channels` attribute.
+    #: the :attr:`source`'s :attr:`~acoular.base.Generator.num_channels` attribute.
```
artpelling, line 2690: are you sure that this is rendered correctly in Shinx?

acoular/tprocess.py
```diff
@@ -2687,10 +2687,22 @@ class TimeConvolve(TimeOut):
...
+    #:
+    #: - If ``False`` (default): Output length is :math:`\\max(L, M)`, where :math:`L` is the
+    #:   kernel length and :math:`M` is the signal length. This mode keeps the output length
+    #:   equal to the longest input (different from NumPy's ``mode='same'``, since it does not
+    #:   pad the output).
+    #: - If ``True``: Output length is :math:`L + M - 1`, returning the full convolution at
+    #:   each overlap point (similar to NumPy's ``mode='full'``).
+    #:
+    #: Default is ``False``.
+    extend_sample = Bool(False, desc='Extend result sample')
```
artpelling: `extend_signal` is a better name for the trait. Also, note that we do not use the `desc=` keyword anymore. The Sphinx doc-comment is sufficient.

acoular/tprocess.py
```diff
@@ -2776,25 +2788,29 @@ def result(self, num=128):
         Yields
         ------
         :obj:`numpy.ndarray`
-            A array of shape (``num``, :attr:`~acoular.base.SamplesGenerator.num_channels`),
-            where :attr:`~acoular.base.SamplesGenerator.num_channels` is inhereted from the
+            An array of shape (``num``, :attr:`~acoular.base.Generator.num_channels`),
+            where :attr:`~acoular.base.Generator.num_channels` is inherited from the
```
artpelling, line 2792: same here with `Generator` ...

acoular/sources.py
jtodev: The new default for `TimeConvolve` is `extend_sample = False`, so the new result is shorter. I could of course set `extend_example` to `True` and keep the old result.

acoular/tprocess.py
jtodev: Yes, `num_channels` is a property defined in the `Generator` class. The old reference did not render correctly.

acoular/tprocess.py
jtodev: All right. I will change it.

PR #618, opened 2026-06-05
Organizational level `ruff` configuration

artpelling:
As @adku1173 suggested. I gave it some more thought and I think this could be viable until a first-class solution is released. See https://github.com/astral-sh/ruff/issues/12352

The only drawback is the hardcoded path to `.venv/share/ruff/ruff.acoular.toml` which soft-forces a `uv`-based workflow. But since this is the way anyway it is the way I'd say ...


esarradj:
Unfortunately as long as the clash between numpy multi threading and numba is not fixed, Acoular needs to be imported before numpy or it will run only one threaded and very slow. Even if ruff doesn't like it.

artpelling:
> Unfortunately as long as the clash between numpy multi threading and numba is not fixed, Acoular needs to be imported before numpy or it will run only one threaded and very slow. Even if ruff doesn't like it.

I see. As this is not really what this PR is about, I will revert the import order changes and add `noqa` directives to it to safeguard this for the future.

Regarding the underlying issue, I have started investigating and might have found a slightly better "solution". Will propose it when I tested it.

artpelling:
The `noqa` directives did not work consistently. Even better now, I've added an `isort` config that enforces a custom import order, see https://github.com/acoular/ci/blob/c05d078843d6b9dd7232684651c284a252aabdbd/ruff.acoular.toml#L36-L40

I am quite pleased with the way this acts. Thoughts @esarradj @adku1173 ?

artpelling:
Once we've locked down all linting rules, I will create a release in the CI repo and pin that for the `acoular-ruff` dependency in `pyproject.toml`. That way, we can ensure all Acoular packages use the exact same linting config and ruff version with a central configuration. 

So if we decide to merge this PR, this would be a great opportunity to revise the linting ruleset and maybe activate more rules (can be overridden/ignored in the individual packages to allow gradual adoption).

adku1173 (dismissed):

artpelling:
@jtodev There are 61 linting violations left that need some minor refactoring, could you please look at those and fix them?

```
13	SIM115 	open-file-with-context-handler
 7	W505   	doc-line-too-long
 3	INP001 	implicit-namespace-package
 3	PLC0415	import-outside-top-level
 3	S318   	suspicious-xml-mini-dom-usage
 2	ARG001 	unused-function-argument
 2	ARG002 	unused-method-argument
 2	B018   	useless-expression
 2	S102   	exec-builtin
 2	S301   	suspicious-pickle-usage
 2	S310   	suspicious-url-open-usage
 2	S324   	hashlib-insecure-hash-function
 2	S607   	start-process-with-partial-path
 2	TID253 	banned-module-level-imports
 2	TRY002 	raise-vanilla-class
 1	A001   	builtin-variable-shadowing
 1	B007   	unused-loop-control-variable
 1	B023   	function-uses-loop-variable
 1	BLE001 	blind-except
 1	E722   	bare-except
 1	FIX004 	line-contains-hack
 1	N813   	camelcase-imported-as-lowercase
 1	PLW1508	invalid-envvar-default
 1	S101   	assert
 1	S110   	try-except-pass
 1	S307   	suspicious-eval-usage
 1	S603   	subprocess-without-shell-equals-true
Found 61 errors.
```

adku1173:
@jtodev @artpelling, is this PR ready? I suggest to merge the current changes and fix the remaining rules in another PR.

jtodev:
> @jtodev @artpelling, is this PR ready? I suggest to merge the current changes and fix the remaining rules in another PR.

If there are no questions regarding my changes, I consider it finished for now.

adku1173 (approved):

PR #636, opened 2026-07-16
Add notch filters for tonal noise suppression

stebuech:
Adds four notch filter classes to `acoular.tprocess` for suppressing tonal
components — typically the blade-passing frequency (BPF) of a propeller and its
harmonics — from array measurements. The motivating case is a drone carrying its
own microphone array, where rotor ego-noise masks the sources of interest.

Acoular currently has no notch filter, so this is a greenfield addition with no
back-compatibility constraints. The implementation follows Harvey (2019).

## What's added

| Class | Purpose |
| --- | --- |
| `NotchFilter` | Second-order IIR notch at a fixed frequency. Subclasses `Filter` and provides a single `(1, 6)` SOS section via `_get_sos`, so streaming and per-channel state are inherited. |
| `AdaptiveNotchFilter` | Tracks a time-varying tone, either from an external frequency source (e.g. RPM data) or autonomously via LMS. |
| `ZeroPhaseNotchFilter` | Forward-backward filtering, preserving the phase relations beamforming depends on. |
| `CascadeNotchFilter` | S × M bank of notches for harmonic series of one or several sources. |

The classes build on each other (`NotchFilter` → `AdaptiveNotchFilter` →
`ZeroPhaseNotchFilter`; `CascadeNotchFilter` composes a bank of them).

Supporting Numba kernels go into `acoular.tfastfuncs` with explicit type
signatures, matching the existing style there.

## Notes on correctness

- For a fixed notch frequency in `batch_mode`, `ZeroPhaseNotchFilter` is
  numerically identical to `scipy.signal.filtfilt` (max deviation `0.0`).
- The `(1 - r) * fs / pi` bandwidth approximation was validated against
  `freqz`: at 25.6 kHz and 200 Hz it predicts 81.5 Hz for `pole_radius=0.99`
  against a measured 81.1 Hz.
- Notch bandwidth scales with the sampling frequency, and a narrow notch settles
  slowly (`7 / |ln r|` samples). Both are documented, as they are easy to get
  wrong.

## Documentation and tests

- New user guide section *Notch filtering of tonal noise*.
- New gallery example covering all four classes (runs in ~0.65 s).
- 110 tests under `tests/unittests/notch/`.
- Full `tests/unittests/`: 1170 passed, 0 failed. `ruff check` and
  `ruff format --check` pass repo-wide. Docs build succeeds with no new
  warnings.
- References added to `literature.bib`; docstrings use `:cite:`.

## Points I'd like feedback on

1. **`AdaptiveNotchFilter.learned_frequencies`** — I added this public property
   because auto mode was otherwise unreadable without touching private state,
   and `CascadeNotchFilter` already had the equivalent. It is new public API, so
   it deserves a look. (Both hold the last block only.)
2. **`ZeroPhaseNotchFilter` warns during normal streaming** when the lookahead
   buffer is shorter than the settling time. Because tests treat warnings as
   errors, `ZeroPhaseNotchFilter` needed a dedicated generator case with
   `batch_mode=True` in `tests/cases/test_generator_cases.py`. Should a library
   class warn here at all, or is documenting the trade-off enough?
3. **`find_spectral_peaks`** is currently public in `tprocess` but not in the
   autosummary. Make it private instead?
4. **`CascadeNotchFilter.joint_lms` defaults to `False`** (independent per-stage
   tracking). Setting it to `True` selects the joint optimisation from the
   thesis, which locks harmonic *m* to *m* times the fundamental — a strong
   constraint that helps only when it actually holds.

Two scoped `per-file-ignores` were added to `pyproject.toml`: `N806` for
`tfastfuncs.py` (DSP convention `N, K, S, M`) and a set of style rules for the
ported tests. Happy to clean those up instead if preferred.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ docs/sg_execution_times.rst
 docs/api_ref/generated/
 docs/auto_examples/**/*
 docs/_static/switcher.json
+
+.idea/
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -44,6 +44,9 @@ authors:
     family-names: "Schumacher"
     affiliation: "Deutsches Zentrum für Luft- und Raumfahrt (DLR)"
     orcid: "https://orcid.org/0000-0001-8349-2570"
+  - given-names: "Steffen"
+    family-names: "Büchholz"
+    affiliation: "Department of Engineering Acoustics, TU Berlin"
 identifiers:
   - description: "acoular: all versions"
     type: "doi"
--- a/acoular/__init__.py
+++ b/acoular/__init__.py
@@ -109,7 +109,9 @@ from .tbeamform import (
     IntegratorSectorTime,
 )
 from .tprocess import (
+    AdaptiveNotchFilter,
     AngleTracker,
+    CascadeNotchFilter,
     ChannelMixer,
     Filter,
     FilterBank,
@@ -118,6 +120,7 @@ from .tprocess import (
     FiltOctave,
     MaskedTimeOut,
     Mixer,
+    NotchFilter,
     OctaveFilterBank,
     SpatialInterpolator,
     SpatialInterpolatorConstantRotation,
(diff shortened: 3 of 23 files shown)


artpelling:
I will convert this to a draft PR, until we've discussed how to move forward.

PR #640, opened 2026-07-18
Reuse Examples in user guide

artpelling:
To avoid untested/drifting documentation, all code in the user guide should be tested against the latest Acoular version. I think the best way is to only allow including code snippets from `examples/` in the user guide. For code that is not yet available, an example should be added in the future. I've updated the current guide to follow this idea. All code blocks are included with `..literalinclude`

### DIFF
--- a/docs/user_guide/caching.rst
+++ b/docs/user_guide/caching.rst
@@ -13,6 +13,9 @@ Lazy evaluation decides when a calculation is needed. Caching decides whether
 Acoular can reuse data from an earlier calculation. This chapter explains the
 file-based caching controls.
 
+The code snippets below are taken from
+:doc:`../auto_examples/introductory_examples/example_caching`.
+
 What Is Cached
 --------------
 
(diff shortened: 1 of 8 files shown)


esarradj (commented):
I understand the intention that if we reuse the examples in the user guide it would stay up to date. However, what happens if for some reason we change the examples? We would possibly end up with nonsensical (or completely missing) code snippets in the user guide. So I am not sure if this is really an improvement.

Apart from this I fixed the many '# %%' lines that appeared in the code snippets.

artpelling:
> I understand the intention that if we reuse the examples in the user guide it would stay up to date. However, what happens if for some reason we change the examples? We would possibly end up with nonsensical (or completely missing) code snippets in the user guide. So I am not sure if this is really an improvement.

I am heavily against untested code in the documentation, we still have a lot of documenting debt and keep discovering out-of-date parts of the docs. I am not claiming that this PR resolves this problem entirely but the chance for drift is greatly reduced.

Before, any change of code could break the user guide. Now only a change of examples can do so (we rarely touched those in the past). It is also not too hard to double check the user guide if an example needs changing. I wouldn't add a test for this but we could add a comment header with a reminder in which parts of the user guide an example is embedded in.

esarradj (approved):

PR #642, opened 2026-07-21
Fix Average to save memory in its source processing

esarradj:
When `Average.result` fetches its data from its source it does so using possibly very large block sizes. This could overwhelm the processing in e.g. `BeamformerTimeTrajSq` by allocating too much memory (and eventually many many GB - more than the computer might have). This is fixed by just pulling as many samples as needed to average over.

### Checklist

### DIFF
--- a/acoular/process.py
+++ b/acoular/process.py
@@ -159,8 +159,8 @@ class Average(InOut):
         ----------
         num : :class:`int`
             The number of averaged blocks to yield at a time. Each block contains the average over
-            :attr:`num_per_average` time samples or frequency snapshots. The last block may be
-            shorter than the specified size if the remaining data is insufficient.
+            :attr:`num_per_average` time samples or frequency snapshots. The last yielded block
+            may contain fewer than ``num`` averages.
 
         Yields
         ------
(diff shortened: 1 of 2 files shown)


acoular/process.py
```diff
@@ -180,14 +180,34 @@ def result(self, num):
...
+                # create accumulator if not exists
+                if out is None:
+                    out = np.empty((num, nc), dtype=avg.dtype)
+                # add one output sample
+                out[outnum] = avg
+                outnum += 1
+                # block complete -> yield it and create new one
+                if outnum == num:
+                    yield out
+                    out = np.empty_like(out)
```
artpelling:
```suggestion
                    out = None
```

may save one allocation 🤡

esarradj: No, it does not. L199 will only get executed when out is None. Actually, assigning None is one (tiny) allocation more, because then L199 will be executed with a second allocation.

artpelling: But it would save the empty allocation in the final block :D

esarradj: ok, ok.

artpelling (approved):

Issue #27, opened 2019-11-22
[Feature Proposal] PolarGrid


beasteers:
You have `RectGrid`, but it seems like defining grids in polar coordinates might be even more useful.

gherold:
Hi Bea,
thanks for the suggestion. I wouldn't say that polar coordinates are _more_ useful (really depends on what you want to do), but I agree that flexible grid capabilities (e.g. loading arbitrary focus grids from files) would cover more use cases. This actually has been on our to-do list for quite some time and will be considered for the next release.
Cheers,
Gert

beasteers:
Sorry I didn't mean to imply that RectGrid wasn't useful, I was just having trouble conceptualizing where I should orient my rectangular grid wrt the microphones and was thinking that defining it at a constant radius from the mic array origin would make more sense (in my use case).

Just a thought (as a generalization/simplification of grids), instead of having `Grid` and `3DGrid`, you could potentially do:
```python
RectGrid(x=(-3.0, 3.0), y=-3.0, z=(-4, 4), increment=0.3) # 2d
RectGrid(x=(-3.0, 3.0), y=(-4, 4), z=1, increment=0.3) # 2d
RectGrid(x=(-3.0, 3.0), y=(-4, 4), z=(-4, 4), increment=0.3) # 3d
RectGrid(x=(-3.0, 3.0), y=2, z=1, increment=0.3) # 1d
```
Which covers a 1D "grid" as well.

Great, glad to hear! thanks for the response, and very nice package! It's nice to see good high level acoustics libraries out there.

artpelling:
Hmm, I don't think this should be a problem. After all, `RectGrid` is just one possible implementation of `Grid`. Maybe looking at implementations of something less structured such as `ImportGrid` or `MergeGrid` could give you some ideas how to implement it. 
 > For this, we need to discuss if the size attribute is a description of memory space or the number of grid points. In the RectGrid, this is the same (which explains why it is in the current ambiguous state ...)

I am not sure what you mean by memory space. `size` should just reflect the number of points in the grid. Maybe that attribute should have a more descriptive name, something starting with `num_` perhaps.

PR #646, opened 2026-07-27
[Feature Proposal] PolarGrid #27

MSzBerlin:
Still Work in Progress:

Implemented PolarGrid(Grid) class similar to Gert's implementation.

Valid parameter ranges are enforced with private validate method.

In case of r_min = 0, duplicate points at origin are removed from pos array.

BUT at the moment some pytests fail (mainly subdomain and sector integration) since a polar grid including the origin (r=0) has an uneven number of grid points and the definition of "shape" needs to be thought about because in this case it is no longer (nrsteps, nphisteps)...

I have to solve this, when I am back in office.

### DIFF
--- a/acoular/__init__.py
+++ b/acoular/__init__.py
@@ -60,6 +60,7 @@ from .grids import (
     LineGrid,
     MergeGrid,
     MultiSector,
+    PolarGrid,
     PolySector,
     RectGrid,
     RectGrid3D,
--- a/acoular/grids.py
+++ b/acoular/grids.py
@@ -512,9 +512,7 @@ class PolarGrid(Grid):
     nphisteps = Property()
 
     #: A unique identifier for the grid, based on its properties. (read-only)
-    digest = Property(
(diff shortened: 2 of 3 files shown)


esarradj:
I am not really sure about what the `PolarGrid` should accomplish, and I think before continuing we need to agree on this. 

Presently, the implementation defines a `"RectGrid` in polar coordinates. The biggest problem with this is that we do not get an equally distributed density of grid points (this escalates a radius=0, where we have a number of colocated grid points). Another problem is the periodicity in phi that needs to handled if we do allow limits for phi. This requires a lot of extra limit testing and validation and would be good idea only if its worth the cost.

I think, what we really need are a grids with a full circular shape and with a ring shape. Are there any relevant use cases for a half-circle or similar, i.e. do we need the limits for phi at all? Also, do we need a regular phi spacing? Or could we just have a different number of grid points along the circle for each radius? The latter would help to have similar spacing of the grid points everywhere. The downside would be to have not a perfect symmetry in the grid then.

@artpelling @adku1173  @gherold : any thoughts on this?

PR #648, opened 2026-07-30
extends TimeSamplesAIAABenchmark class …

gherold:
… to be able to load data stored in older revisions of the file format (transposed). It will check which dimension of the data array is larger and assume that this represents the time samples, while the lower number is channels. If data has to be transposed, a warning will be shown.

### Checklist


acoular/aiaa/aiaa.py
```diff
@@ -55,18 +57,103 @@ class TimeSamplesAIAABenchmark(TimeSamples):
...
+                )
+                self.num_channels, self.num_samples = self.data.shape
+            else:
+                self.num_samples, self.num_channels = self.data.shape
+
     def _load_timedata(self):
-        """Loads timedata from .h5 file. Only for internal use."""
+        """Loads timedata from :attr:`.h5 file<file>`. Only for internal use."""
         self.data = self._h5f.get_data_by_reference('MicrophoneData/microphoneDataPa')
         self.sample_freq = self._h5f.get_node_attribute(self.data, 'sampleRateHz')
```
esarradj, line 107: there is a node attribute 'sampleCount' that can be used to determine whether `self.data` is transposed

acoular/aiaa/aiaa.py
```diff
@@ -55,18 +57,103 @@ class TimeSamplesAIAABenchmark(TimeSamples):
...
+        if self._data_transposed:
+            return self.data[0, :].sum()
+        return self.data[:, 0].sum()
+
+    @observe('data')
+    def _load_shapes(self, event):  # noqa: ARG002
+        # Set :attr:`num_channels` and :attr:`num_samples` from data.
+        if self.data is not None:
+            data_shape = self.data.shape
+            self._data_transposed = data_shape[0] < data_shape[1]
```
esarradj: The information about the number of channels and samples is stored in the file as H5 attributes already. Guessing around may have unwanted side effects.

esarradj (commented):
I would suggest to drop `_load_shapes` and guessing the dimensions of `data`, but instead implement this in `_load_timedata` using the H5 attribute.

acoular/aiaa/aiaa.py
gherold: _load_shapes follows the structure of the parent class, so is kept. The "sampleCount" attribute is now read to determine whether the data is interpreted as transposed

acoular/aiaa/aiaa.py
gherold: Guessing is now partially replaced by reading the "sampleCount" attribute (if possible).

esarradj (approved):

PR #650, opened 2026-08-04
avoid copying avg

artpelling:
This will avoid instantiating `avg` and copying that into `out`. Also, I removed some nesting of if clauses, hoping that its easier to grasp.

### DIFF
--- a/acoular/process.py
+++ b/acoular/process.py
@@ -201,14 +201,14 @@ class Average(InOut):
             # create accumulator if not exists
             if out is None:
                 out = np.empty((num, nc), dtype=dtype)
-                # add one output sample
-                temp.mean(axis=0, out=out[outnum])
-                outnum += 1
-                # block complete -> yield it and create new one
-                if outnum == num:
-                    yield out
-                    out = None
-                    outnum = 0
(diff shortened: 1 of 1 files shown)


acoular/process.py
```diff
@@ -187,18 +187,22 @@ def result(self, num):
...
-            if ns == nav:
-                avg = temp.mean(axis=0)
-                # create accumulator if not exists
-                if out is None:
-                    out = np.empty((num, nc), dtype=avg.dtype)
+            if ns != nav:
+                continue
+            # determine dtype of the mean once
+            if dtype is None:
+                dtype = np.empty((), dtype=temp.dtype).mean().dtype
```
esarradj: would it be possible for the mean to have a different dtype than temp?

artpelling: if temp was int, mean would be float I think. not really realistic, we could also skip it...

artpelling: @esarradj so I just use temp.dtype? (one more thumb pls)

esarradj: 👍

artpelling: :+1:

PR #651, opened 2026-08-04
Speedup of package import

esarradj:
Acoular import can be slow.

**Summary**

- Speed up `import acoular` by lazily loading optional/heavy modules.
- Add shared `lazyjit` helper in `acoular.internal` and apply it to all Numba-jitted functions.
- Lazy-load `sounddevice` and scikit-learn solver classes

**Details**

- `sounddevice` is imported only when `SoundDeviceSamplesGenerator` is used.
- `sklearn` imports in `fbeamform.py` now happen only inside solver paths that need them.
- All direct Numba decorators in the package are now wrapped as `@lazyjit(nb...)`, delaying Numba dispatcher creation until first function call.

### Checklist


artpelling:
Ohh nice, this is something I wanted to do for a long time!

artpelling:
Looks good. Just a side note: Not all decorators implement `FAST_OPTION` and `CACHED_OPTION` consistently. Is this correct?

esarradj:
Yes indeed. This was only implemented in fastFuncs.py 
While jitted functions need (heavy) reworking, I would prefer to leave this problem alone for now.

artpelling:
> Yes indeed. This was only implemented in fastFuncs.py While jitted functions need (heavy) reworking, I would prefer to leave this problem alone for now.

I also think we should not touch this here, it might be worthwile to add these options inside `lazyjit` and remove them as kwargs so they cannot be forgotten by accident.

PR #652, opened 2026-08-04
fix conda yml

artpelling:
Conda Release CI was always failing, I think I found the culprit now.


esarradj:
I guess we don't need to update What's new because this is just the ci

esarradj (approved):

Issue #628, opened 2026-07-15
Subissue 1: Extract CMF problem assembly helpers


adku1173:
# Subissue 1: Extract CMF problem assembly helpers

**Issue type:** Feature request / refactoring subissue  
**Labels:** `enhancement`, `new feature`  
**Type:** AFK  
**Blocked by:** None — can start immediately  
**Parent:** #632

## Feature Description

Refactor the existing Covariance Matrix Fitting (CMF) implementation so that the construction of the CMF inverse problem is separated from the solver execution path.

At the moment, the CMF `_calc` implementation on `master` assembles the full problem inline: it builds the sensing matrix / dictionary from the steering transfer functions, selects the reduced CSM entries, vectorizes the CSM, applies the real/imaginary representation, and then runs the selected optimization method. This makes it harder to reuse the assembled CMF problem with other solvers in later changes. See the code references below for durable links to the current implementation and the branch that can be used as guidance.

This subissue should be a behavior-preserving refactoring only. It should introduce internal helper structure comparable to the current `clean_solver` branch guidance, but it must not introduce a new public solver interface yet.

The refactoring should add internal helpers for:

- computing and reusing the reduced matrix/vector indices used for the CMF problem,
- building the real-valued CMF sensing matrix / dictionary for a given frequency,
- vectorizing the CSM with the same reduction and real/imaginary representation.

The existing CMF solver behavior, scaling, method selection, and public API should remain unchanged.

## Use Case

This is the first step toward making alternative solvers usable for CMF, and later potentially for other beamformers. By separating problem assembly from solver execution, later changes can pass the same CMF sensing matrix and vectorized CSM to different optimization backends without duplicating or modifying the core assembly logic.

For users, this should not change results or usage. For maintainers and future contributors, it makes the CMF implementation easier to test, reason about, and extend.

## What to build

Refactor the CMF implementation so that `_calc` delegates CMF problem assembly to internal helpers while preserving the existing behavior.

The implementation should follow this shape conceptually:

- introduce an internal cached/index helper for the reduced CSM/Kronecker-product entries,
- introduce an internal helper that returns the CMF sensing matrix / dictionary for one frequency,
- introduce an internal helper that returns the vectorized CSM for one CSM matrix,
- update `_calc` to use those helpers before running the existing optimization methods.

[...]

PR #661, opened 2026-08-06
Cmf problem assembly

Narges825:
### Description

Add msm_indices, _calc_sensing_matrix, and _vectorize_csm to
BeamformerCMF, and update _calc to delegate to them instead of
inlining the sensing-matrix/CSM-vectorization logic. Solver
dispatch is unchanged. Verified numerically unchanged against
the existing CMF regression snapshots for both r_diag settings.

### Checklist

- [ ] I have read the [Contributing](https://www.acoular.org/contributing/index.html) section.
- [ ] I have updated the [What's new](https://github.com/acoular/acoular/blob/master/docs/news/index.rst) section explaining my changes.


acoular/fbeamform.py
```diff
@@ -1672,6 +1672,42 @@ def _validate(self, event):  # noqa: ARG002
             )
             raise ImportError(msg)
 
+    #: Cached index masks (triangular + real/imag selection) for the CMF problem.
+    msm_indices = Property(depends_on=['freq_data.num_channels', 'steer.digest', 'r_diag'])
```
adku1173: I think msm_indices should be renamed to `_csm_indices` so that it is a private attribute and the variable name is somewhat clearer. 

Why does it depend on the steering vector digest? I think `freq_data.num_channels` and `r_diag` should be enough. If `steer.digest` is included, recalculation would be triggered too often.

acoular/fbeamform.py
```diff
@@ -1672,6 +1672,42 @@ def _validate(self, event):  # noqa: ARG002
...
+    def _get_msm_indices(self):
+        """Indices for the reduced Kronecker product / vectorized CSM entries."""
+        nc = self.freq_data.num_channels
+        ind = np.reshape(np.tril(np.ones((nc, nc))), (nc * nc,)) > 0
+        ind_im0 = (np.reshape(np.eye(nc), (nc * nc,)) == 0)[ind]
+        if self.r_diag:
+            ind_reim = np.hstack([ind_im0, ind_im0])
+        else:
+            ind_reim = np.hstack([np.ones(np.size(ind_im0)) > 0, ind_im0])
+            ind_reim[0] = True  # why this ?
```
adku1173: Remove the comment

```suggestion
            ind_reim[0] = True
```

acoular/fbeamform.py
```diff
@@ -1672,6 +1672,42 @@ def _validate(self, event):  # noqa: ARG002
...
+        return ind, ind_reim
+
+    def _calc_sensing_matrix(self, f):
```
adku1173: I would suggest to rename the function to `_build_dictionary` because nothing is calculated here

.gitignore
```diff
@@ -31,3 +31,5 @@ docs/sg_execution_times.rst
 docs/api_ref/generated/
 docs/auto_examples/**/*
 docs/_static/switcher.json
+xxx.py
```
adku1173: Can we remove this from the gitignore?

pyproject.toml
```diff
@@ -182,6 +182,7 @@ banned-module-level-imports = ["matplotlib", "pylops", "sounddevice"]
 ]
 "tests/*" = ["FIX", "PLC0415", "PT", "RUF", "S101", "TRY"]
 "tests/cases/*" = ["N802"]
+"tests/unittests/test_fbeamform.py" = ["N806"]
```
adku1173: I am against excluding N806 rule, whenever we can also simply rename the variables. Maybe use something like `dictionary_full` etc. `test_fbeamform.py` is also slightly too general for the unittests. Maybe better use `test_cmf_problem_setup.py` or `test_cmf_assembly.py`

adku1173 (changes requested):
Thank you @Narges825 :). I have some comments before we finally merge these changes.

acoular/fbeamform.py
Narges825: Good catch! I’ve renamed `msm_indices` to `_csm_indices`. I’ll also remove `steer.digest` from the dependencies since the getter only uses `freq_data.num_channels` and `r_diag`.

acoular/fbeamform.py
Narges825: I think the whole line can be removed, since ind_reim[0] is already True by construction.

acoular/fbeamform.py
Narges825: Done!

.gitignore
Narges825: Yes, i did.

pyproject.toml
Narges825: Thanks! I’ve made the suggested changes. I renamed the variables instead of excluding N806 and renamed `test_fbeamform.py` to `test_cmf_problem_setup.py`.

adku1173 (approved):

adku1173:
closes #628

PR #662, opened 2026-08-11
Compute the CSM eigendecomposition only once for eva and eve

### LINKED ISSUE #111: Unnecessary double calculation of eigvals and eigvecs.
When calculating eigenvalues and eigenvectors, the function is called twice (once for each calculation, see lines 245-251 in [spectra.py](acoular/spectra.py#L245) ). Better would be to perform the calculation only once for both if one of the two is needed and cache the results in the corresponding traits. Might be connected to #110.

siavashzaid:
### Description

`PowerSpectra.eva` and `PowerSpectra.eve` each triggered their own call to
`calc_ev()`, so accessing both ran `numpy.linalg.eigh` over every frequency
twice, even though a single decomposition yields both results.

**Performance fix**

The eigendecomposition now lives in an internal, cached property `_evd` holding
the `(eigvals, eigvecs)` tuple. Its getter is decorated with
`@property_depends_on(['digest'])`, so the decomposition runs once and is
invalidated whenever the digest changes. Both public traits read from it, so accessing eigenvalues and
eigenvectors costs one decomposition instead of two.

**Renaming and deprecation**

While touching this code, the cryptic trait names were replaced:

* `eva` → `eigvals`, `eve` → `eigvecs`. The old names remain available through
  the `deprecated_alias` decorator and raise a `DeprecationWarning`.
* `calc_ev()`, `calc_eva()` and `calc_eve()` are deprecated. They still work and
  forward to the new traits. The eigendecomposition is now computed on access of
  `eigvals` / `eigvecs` so no public method for triggering it is needed.
* All internal callers / mentions were updated accordingly.

### Reference issue
Closes #111 

### Checklist


acoular/spectra.py
```diff
@@ -261,6 +261,11 @@ class PowerSpectra(BaseSpectra):
     #: :attr:`~acoular.spectra.BaseSpectra.num_channels`. (read-only)
     eve = Property()
 
+    # Eigenvalues and eigenvectors of the CSM as a tuple ``(eva, eve)``, computed in a single
+    # decomposition and cached so that :attr:`eva` and :attr:`eve` share one call to
+    # :meth:`calc_ev`. For internal use only.
+    _ev = Property()
```
artpelling:
```suggestion
    _evd = Property()
```
EVD would be a better name since it is the decomposition

acoular/spectra.py
artpelling, line 428: the `calc_eva` and `calc_eve` methods should be deprecated. We should introduce new properties `eigvals` and `eigvecs` with getters that access the shadow trait.

artpelling (changes requested):

acoular/spectra.py
```diff
@@ -339,6 +348,22 @@ def _get_digest(self):
...
 
+    @property_depends_on(['digest'])
+    def _get__evd(self):
+        # Eigendecomposition of the CSM for all frequencies as a tuple ``(eigvals, eigvecs)``.
+        # Cached so that :attr:`eigvals` and :attr:`eigvecs` need only one decomposition.
+        if self.precision == 'complex128':
+            eigvals_dtype = 'float64'
+        elif self.precision == 'complex64':
+            eigvals_dtype = 'float32'
+        #        csm = self.csm #trigger calculation
```
artpelling:
```suggestion
```

acoular/fbeamform.py
```diff
@@ -2405,8 +2405,8 @@ def func(xy):
                 # store result for position
                 self._gpos[:, i1] = oR['x']
                 # store result for level
-                self._ac[i, i1] = eva[n] / num_channels
-                # print(oR['x'],eva[n]/num_channels,oR)
+                self._ac[i, i1] = eigvals[n] / num_channels
+                # print(oR['x'],eigvals[n]/num_channels,oR)
```
artpelling:
```suggestion
```

acoular/spectra.py
```diff
@@ -181,6 +183,7 @@ def _get_source_data(self):
             pos -= bs
 
 
+@deprecated_alias({'eva': 'eigvals', 'eve': 'eigvecs'}, read_only=True)
```
artpelling:
```suggestion
@deprecated_alias({'eva': 'eigvals', 'eve': 'eigvecs'}, read_only=True, removal_version='27.04')
```

artpelling (changes requested):
Looks good, just some minor things.

esarradj (dismissed):

artpelling (approved):

artpelling (approved):

PR #663, opened 2026-08-13
fix and improve finite/infinite `TimeOut`

artpelling:
### Description

While investigating `spectacoular` bugs, I noticed several inconsistencies regarding inifnite streams with `num_samples=-1`. Quite many cases did not support this but would not throw an error. I've added a new mixin class `FiniteTimeOut` that can handle these checks and fixed edge-cases. I've also fixed `MaskedTimeOut` to handle infinite streams (if `stop` is `None`) which means that we can remove `MaskedChannels` from `spectacoular`.
I found a bug in `WriteH5`, where the final block of the written file did not match the yielded one.

I've concentrated on just fixing and not changing behaviour. Several question marks regarding the overall design remain, though: It makes no sense to me that `PowerSpectra` cannot handle infinite streams. Especially since there are a couple of workaround classes and mechanics around. I strongly believe that we are should refactor this. Also, `WriteH5` is awkward because it can take an infinite stream with `num_write_samples`. At least, `WriteWAV` should do the same or both should just be able to write finite streams.

### Checklist

- [ ] I have read the [Contributing](https://www.acoular.org/contributing/index.html) section.
- [ ] My branch is up-to-date with the *master* branch of the [Acoular repository](https://github.com/acoular/acoular).
- [ ] My changes fulfill the [Code Quality](https://www.acoular.org/contributing/quality.html) standards.
- [ ] I have updated the [What's new](https://github.com/acoular/acoular/blob/master/docs/news/index.rst) section explaining my changes.

### DIFF
--- a/tests/unittests/test_indefinite.py
+++ b/tests/unittests/test_indefinite.py
@@ -41,7 +41,7 @@ def test_streaming_blocks_preserve_indefinite_length(monkeypatch):
 def test_time_convolve_streams_indefinite_source():
     source = ac.MaskedTimeOut(source=IndefiniteSamples(), start=3)
     convolve = ac.TimeConvolve(source=source, kernel=np.array([1.0]))
-    np.testing.assert_array_equal(next(convolve.result(4)).ravel(), [3.0, 4.0, 5.0, 6.0])
+    np.testing.assert_allclose(next(convolve.result(4)).ravel(), [3.0, 4.0, 5.0, 6.0], atol=1e-15)
 
 
 @pytest.mark.parametrize(
--- a/tests/unittests/test_indefinite.py
+++ b/tests/unittests/test_indefinite.py
@@ -26,7 +26,7 @@ def test_streaming_blocks_preserve_indefinite_length(monkeypatch):
 
     masked = ac.MaskedTimeOut(source=source, start=3)
     assert masked.num_samples == -1
-    assert next(masked.result(4)).shape == (1, 1)
+    np.testing.assert_array_equal(next(masked.result(4)).ravel(), [3., 4., 5., 6.])
 
     average = ac.Average(source=source, num_per_average=2)
     assert average.num_samples == -1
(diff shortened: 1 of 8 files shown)


tests/unittests/test_indefinite.py
```diff
@@ -0,0 +1,85 @@
...
+            yield np.arange(offset, offset + num)[:, np.newaxis]
+            offset += num
+
+
+def test_streaming_blocks_preserve_indefinite_length(monkeypatch):
+    source = IndefiniteSamples()
+
+    masked = ac.MaskedTimeOut(source=source, start=3)
+    assert masked.num_samples == -1
+    assert next(masked.result(4)).shape == (1, 1)
```
esarradj: I would expect (4,1).

esarradj: but the test does NOT fail ??

tests/unittests/test_indefinite.py
```diff
@@ -0,0 +1,85 @@
...
+    monkeypatch.setattr(ac.config, 'global_caching', 'all')
+    cache = ac.Cache(source=source)
+    assert next(cache.result(4)).shape == (4, 1)
+    assert cache._h5f is None
+
+
+def test_time_convolve_streams_indefinite_source():
+    source = ac.MaskedTimeOut(source=IndefiniteSamples(), start=3)
+    convolve = ac.TimeConvolve(source=source, kernel=np.array([1.0]))
+    np.testing.assert_array_equal(next(convolve.result(4)).ravel(), [3, 4, 5, 6])
```
esarradj: 3., 4., 5., 6.

esarradj (changes requested):

tests/unittests/test_indefinite.py
artpelling: i made an error in the loop. I've reworked it and intertwined properly with the old finite-time result path

tests/unittests/test_indefinite.py
esarradj: This test still fails in some cases because of slight numerical deviations when the integers get compared to floats.

artpelling: The problem is not the numbers but Windows. Will need to match based on tolerance it seems

esarradj (approved):

